PR #40, opened 2019-11-28
add Additional K8s Objects section

fshot:
Added a basic rundown of the various K8s object specs relevant to KOTS apps.


content/vendor/additional-objects/_index.md
```diff
@@ -0,0 +1,11 @@
+---
+date: "2019-11-27T00:00:00Z"
+lastmod: "2019-11-27T00:00:00Z"
+title: "Additional Kubernetes Objects"
+aliases: [/vendor/additional-objects]
+weight: "2"
+hideFromList: true
+---
+Elements of a KOTS application lifecycle are controlled by various Custom Resource Definitions included in a KOTS application spec.  
+
```
marccampbell: I don’t know if we have defined a term “KOTS application spec” anywhere. What if this was changed to be “included as part of a KOTS application.”?

content/vendor/additional-objects/_index.md
```diff
@@ -0,0 +1,11 @@
...
+These additional Kubernetes Objects serve as metadata for kots and kotsadm, and don't require any CRD controller code to be deployed to the target cluster.
```
marccampbell: Object should not be capitalized.

metadata for ~kots and kotsadm~. How about “metadata for the KOTS application that can be used by the Admin Console”

The message that they don’t require CRD controller code is accurate but I think we should dig into this more. That doesn’t help me understand how it does work (the kubectl plugins can use them), and actually that there isn’t a controller even created for these. I think the message to convey here is that these custom resources (instead of Kubernetes Objects) are removed before deploying and do not have any prerequisites or require any elevated permissions to install to an existing cluster.

content/vendor/additional-objects/config.md
```diff
@@ -0,0 +1,8 @@
+---
+date: 2019-11-27
+linktitle: "Config"
+title: Config
+weight: 20020
+---
+
+Config object specifications control the presentation of a [custom configuration page](/vendor/config/config-screen/), enabling operators to specify any required or optional application-specific config settings at install time, or update those options later through the kotsadm dashboard.
```
marccampbell: enabling operators — is the operator here the application developer? If so, we should define who this role is.

marccampbell: I’d prefer to stick to using the term Admin Console (capitalized) over “kotsadm”. So this sentence should end with “Admin Console Dashboard” instead of “kotsadm dashboard”

content/vendor/additional-objects/k8s-sig-application.md
```diff
@@ -0,0 +1,8 @@
+---
+date: 2019-11-27
+linktitle: "K8s SIG Application"
+title: K8s SIG Application
+weight: 20030
+---
+
+The [Kubernetes SIG Application](https://github.com/kubernetes-sigs/application#kubernetes-applications) spec provides a standard API for creating, viewing, and managing applications.  A SIG Application spec provides helpful metadata to other Kubernetes tools, and enables important KOTS features such as [links to the deployed application](/vendor/dashboard/open-buttons/#provide-an-application-crd).
```
marccampbell: We should list the fields that are required and try to explain how to use this as part of a KOTS application.

content/vendor/additional-objects/kots-application.md
```diff
@@ -0,0 +1,12 @@
...
+linktitle: "KOTS Application"
+title: KOTS Application
+weight: 20010
+---
+
+The [KOTS Application spec](/v1beta1/application/) allows for the customization of the initial experience (title, icon, release notes) and various operational and diagnostic features (port forwarding, statusInformers, and custom graphs).
+
+With ports specified, the kots CLI can establish port-forwarding, to simplify connections to the deployed application.  When [statusInformers](/vendor/dashboard/application-status/#kots-application-spec) are specified, the dashboard can provide timely feedback when the application deployment is complete and the application is ready for use.
+
+There is some overlap between the [KOTS Application spec](/v1beta1/application/) and the [Kubernetes SIG Application spec](https://github.com/kubernetes-sigs/application#application-objects).  In time, it's likely that the SIG Application spec will grow to include all the necessary metadata to support the full KOTS featureset.  In the meantime, best practice will be to include a complete KOTS Application spec along with a complete SIG Application spec in the release YAML of any KOTS application.
```
marccampbell: is it just best practice or is it required? I think we require that you include both to get buttons on the dashboard

content/vendor/additional-objects/kots-application.md
```diff
@@ -0,0 +1,12 @@
+---
+date: 2019-11-27
+linktitle: "KOTS Application"
+title: KOTS Application
+weight: 20010
+---
+
+The [KOTS Application spec](/v1beta1/application/) allows for the customization of the initial experience (title, icon, release notes) and various operational and diagnostic features (port forwarding, statusInformers, and custom graphs).
```
marccampbell: I’m not sure I agree with the grouping here of “initial experience” and “diagnostic features”.  Let’s just list the features that can be enabled by adding this document type.

content/vendor/additional-objects/preflight.md
```diff
@@ -0,0 +1,9 @@
+---
+date: 2019-11-27
+linktitle: "Preflight"
+title: Preflight
+weight: 20040
+---
+A [Preflight spec](https://troubleshoot.sh/reference/preflights/) makes an application's cluster dependencies explicit, enabling kots and kotsadm to provide operators with clear feedback on any missing requirements or incompatibilities in the target environment before an application is deployed.
```
marccampbell: ~kots and kotsadm` KOTS and the Admin Console

~operators~ Cluster Operators

content/vendor/additional-objects/preflight.md
```diff
@@ -0,0 +1,9 @@
...
+
+Preflight checks can even be run independently of kots and kotsadm, via the [Preflight kubectl plugin](https://troubleshoot.sh/docs/preflight/overview/).
```
marccampbell: Let’s remove ~of kots and kotsadm~.

content/vendor/additional-objects/support-bundle.md
```diff
@@ -0,0 +1,14 @@
+---
+date: 2019-11-27
+linktitle: "Support Bundle"
+title: Support Bundle
+weight: 20050
+---
+
+The Support Bundle diagnostic tool collects and analyzes troubleshooting data from a deployed application.  The tool depends on two Custom Resources.
```
marccampbell: collects, redacts and analyzes

marccampbell: It’s not from a deployed application, it’s from a cluster. A useful way to use the support bundle is if the application won’t deploy.

content/vendor/additional-objects/support-bundle.md
```diff
@@ -0,0 +1,14 @@
...
+
+A [Collector Spec](https://troubleshoot.sh/reference/collectors/overview/) defines the data to be collected and included in the .tar.gz bundle.
+
+An [Analyzer Spec](https://troubleshoot.sh/reference/analyzers/overview/) defines how to interpret and present the collected data to the application operator.
+
+Support Bundle collection and analysis are integrated into the [kotsadm dashboard](/kotsadm/troubleshooting/support-bundle/), and are also available through a [standalone kubectl plugin](https://troubleshoot.sh/docs/support-bundle/collecting/).
```
marccampbell: kotsadm  dashboard -> Admin Console Dashboard

content/vendor/dashboard/application-status.md
```diff
@@ -11,7 +11,7 @@ With minimal additional to the kots.io Application YAML, it is possible to displ
 
 It is necessary to target specific Kubernetes resources for the dashboard to accurately report status. We suggest at least one resource be added. Resources that are currently supported are Deployments, StatefulSets, Services, Ingresses and PersistentVolumeClaims.
 
-## Kots Application Spec
+{{< linked_headline "Kots Application Spec" >}}
```
marccampbell: I don’t understand this. Any reason not to use standard Markdown here? the `linked_headline` is proprietary compared to the ## syntax.

content/vendor/additional-objects/_index.md
fshot: Right on.  How about this to replace the second section: "These additional Kubernetes objects serve as metadata for the KOTS application that can be used by the Admin Console, and by kubectl plug-ins on the client side.  There is no CRD controller code required, and thus no need for elevated permissions or installation of any additional components to the target cluster."

content/vendor/additional-objects/config.md
fshot: "Config object specifications control the presentation of a [custom configuration page](/vendor/config/config-screen/), enabling application developers to specify any required or optional application-specific config settings for the cluster operator to provide at install time, or update those options later through the Admin Console."

content/vendor/dashboard/application-status.md
fshot: You are right, this is not necessary.  I had copied this from help-center, as an expedient for making hover-over anchor links, but this is not necessary, and Hugo will still make anchor links from regular markdown headers.

Removing

PR #57, opened 2020-01-05
Guide for installing a helm chart with kots/adm

dexhorthy:
Added a guide for installing a new helm chart with the kots CLI and kotsadm. There are a few things in here that don't 100% work quite right, but overall I think the workflow is solid enough to give folks a sense of what this will look like.

The biggest piece for me is that the `rebuild` command doesn't exist today, so changes to values.yaml don't actually have any effect.


dexhorthy:
CC also @austinchambers

content/kotsadm/installing/helm-chart.md
```diff
@@ -0,0 +1,123 @@
...
+date: 2019-10-09
+linktitle: "Installing a Helm Chart"
+title: Installing a Helm Chart
+weight: 10030
+draft: false
+---
+
+The most direct and simple way to install a KOTS application to a Kubernetes cluster is to deploy to an existing cluster that contains nodes that can access the Internet. In this scenario, the container images will be pulled from the upstream registries directly.
+
+**Note** Currently the only supported source for helm charts are public helm repositories support by helm out-of-the-box.
```
dexhorthy: I'm not sure if this is true or not, but I could be convinced to omit it either way.

PR #64, opened 2020-01-07
Adding an alias for each live page so that we can change things up without damaging our links

austinchambers:
As a means of making it as easy as possible easy to contribute documentation without worrying too much about where it goes, I'd like to implement the convention of title-based-aliasing. In short, do the following;
1. Add /pagelink/page-title to the Aliases list for the page. (e.g., /pagelink/application-spec). 
2. Ensure that all links to kots.io uses URLS that start with https://kots.io/pagelink/...

This way, we can move/reorganize all documentation pages safely without worrying about breaking links.


austinchambers:
Verified that changing the effective URL doesn't break things. The pagelink continues to work.

content/reference/v1beta1/_index.md
```diff
@@ -4,6 +4,8 @@ lastmod: "2020-01-02"
 title: "KOTS Custom Resources"
 aliases: [/v1beta1/]
```
grantmiller: aliases in in here twice

austinchambers: Fixed.

PR #65, opened 2020-01-07
Adding usage/flags information for the kots cli

austinchambers:
Updated all the kots cli commands. I did my best with what information I had, but it needs technical validation for some commands (notably 'kots upload')

image

### DIFF
--- a/content/kots-cli/download.md
+++ b/content/kots-cli/download.md
@@ -5,4 +5,24 @@ title: kots download
 weight: 90120
 ---
 
-Once an application is running and has the Admin Console (kotsadm) deployed with it, you can run `kots download` to retrieve a copy of the application manifests from the cluster, and store them in a specific directory structure on your workstation.
+Retrieves a copy of the application manifests from the cluster, and store them in a specific directory structure on your workstation. Requires a running KOTS application with the Admin Console (kotsadm).
+
+### Usage
+```bash
+kubectl kots download [app-slug] [flags]
+```
+* _Replace `[app-slug]` with the "slug" of your KOTS application_ 
+* _Provide `[flags]` according to the table below_ 
+
+| Flag                 | Type | Description |
+|:----------------------|------|-------------|
+| `--dest` | string |        the directory to store the application in (default "/Users/austin") |
+| `-h, --help`   |  |           help for download |
+| `--kubeconfig` | string |  the kubeconfig to use (default "/Users/austin/.kube/config") |
+| `-n, --namespace` | string |    the namespace to download from (default "default") |
+| `--overwrite` |   |       overwrite any local files, if present |
+
+### Example
+```bash
+kubectl kots download kots-sentry --namespace kots-sentry --dest ./manifests --overwrite
+```
--- a/content/kots-cli/download.md
+++ b/content/kots-cli/download.md
@@ -16,10 +16,10 @@ kubectl kots download [app-slug] [flags]
 
 | Flag                 | Type | Description |
 |:----------------------|------|-------------|
-| `--dest` | string |        the directory to store the application in (default "/Users/austin") |
+| `--dest` | string |        the directory to store the application in _(defaults to $HOME)_ |
 | `-h, --help`   |  |           help for download |
-| `--kubeconfig` | string |  the kubeconfig to use (default "/Users/austin/.kube/config") |
-| `-n, --namespace` | string |    the namespace to download from (default "default") |
+| `--kubeconfig` | string |  the kubeconfig to use. _(default is $KUBECONFIG. If unset, then $HOME/.kube/config)_ |
+| `-n, --namespace` | string |    the namespace to download from _(default "default")_ |
 | `--overwrite` |   |       overwrite any local files, if present |
 
 ### Example
--- a/content/kots-cli/reset-password.md
+++ b/content/kots-cli/reset-password.md
@@ -11,13 +11,13 @@ If you’ve deployed an application with the admin console, the `kots reset-pass
 ```bash
 kubectl kots reset-password [namespace] [flags]
 ```
-* _Replace `[namespace]` with the namespace where your KOTS application resides_ 
+* _Replace `[namespace]` with the namespace where the admin console and your KOTS application resides_ 
 * _Provide `[flags]` according to the table below_ 
 
 | Flag                 | Type | Description |
 |:----------------------|------|-------------|
 | `-h, --help`   |          |  help for reset-password |
-| `--kubeconfig` | string | the kubeconfig to use (default "/Users/austin/.kube/config") |
+| `--kubeconfig` | string | the kubeconfig to use _(default is $KUBECONFIG. If unset, then $HOME/.kube/config)_ |
 
 
 ### Examples
--- a/content/kots-cli/reset-password.md
+++ b/content/kots-cli/reset-password.md
@@ -6,3 +6,21 @@ weight: 90160
 ---
 
 If you’ve deployed an application with the admin console, the `kots reset-password` command will change the bcrypted password hash in the cluster, allowing you to log in again.
+
+### Usage
+```bash
+kubectl kots reset-password [namespace] [flags]
+```
+* _Replace `[namespace]` with the namespace where your KOTS application resides_ 
+* _Provide `[flags]` according to the table below_ 
+
+| Flag                 | Type | Description |
+|:----------------------|------|-------------|
+| `-h, --help`   |          |  help for reset-password |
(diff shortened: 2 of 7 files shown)


content/kots-cli/download.md
```diff
@@ -5,4 +5,24 @@ title: kots download
 weight: 90120
 ---
 
-Once an application is running and has the Admin Console (kotsadm) deployed with it, you can run `kots download` to retrieve a copy of the application manifests from the cluster, and store them in a specific directory structure on your workstation.
+Retrieves a copy of the application manifests from the cluster, and store them in a specific directory structure on your workstation. Requires a running KOTS application with the Admin Console (kotsadm).
+
+### Usage
+```bash
+kubectl kots download [app-slug] [flags]
```
grantmiller: does this require [namespace] as well

content/kots-cli/upload.md
```diff
@@ -5,4 +5,26 @@ title: kots upload
 weight: 90140
 ---
 
-If you have a copy of an application that was created with kots pull or kots download, you can upload it back to the Admin Console using the `kots upload` command.
+Upload Kubernetes manifests from the local filesystem to a cluster, creating a new version of the application that can be deployed. When you have a copy of an application that was created with `kots pull` or `kots download`, you can upload it back to the Admin Console using the `kots upload` command. 
+
+### Usage
+```bash
+kubectl kots upload [source] [flags]
```
grantmiller: does this also require [namespace]

content/kots-cli/upstream.md
```diff
@@ -5,10 +5,22 @@ title: kots upstream
...
 ---
 
-The `upstream` command interacts with a KOTS application that is installed in an Admin Console.
+The `kots upstream upgrade` fetches the latest version of the upstream application. It is functionality equivalent to clicking the Check For Updates in the Admin Console. 
 
-```shell
-kubectl kots upstream upgrade --namespace <namespace> <app-slug>
+### Usage
+```bash
+kubectl kots upstream upgrade [upstream-uri] [flags]
```
grantmiller: does this also require namespace

grantmiller:
these are great, i do think we should have @marccampbell read through them quickly for technical validation. Also, i think there might be a few other commands that need full pages (ie. version)

austinchambers:
@grantmiller FYI -- I copied the output from the help command, which is why [namespace] wasn't included. I can fix this here, but then we also need to fix the help command (as I know we need to do for upstream).

content/kots-cli/upstream.md
austinchambers: namespace is included in the [flags]. Strictly speaking, not a required argument, so I left it out of the usage in this case.

austinchambers: (Note that the: "kubectl kots upstream --help" only returns the "help for upstream" flag).

content/kots-cli/upload.md
austinchambers: namespace is included in the [flags]. Strictly speaking, not a required argument, so I left it out of the usage in this case.

content/kots-cli/download.md
austinchambers: namespace is included in the [flags]. Strictly speaking, not a required argument, so I left it out of the usage in this case. It's in the example though.

content/kots-cli/download.md
```diff
@@ -5,4 +5,24 @@ title: kots download
...
+```
+* _Replace `[app-slug]` with the "slug" of your KOTS application_ 
+* _Provide `[flags]` according to the table below_ 
+
+| Flag                 | Type | Description |
+|:----------------------|------|-------------|
+| `--dest` | string |        the directory to store the application in (default "/Users/austin") |
+| `-h, --help`   |  |           help for download |
+| `--kubeconfig` | string |  the kubeconfig to use (default "/Users/austin/.kube/config") |
```
marccampbell: The default kubecfg flag is dynamic.  If there is a $KUBECONFIG envvar we use that, else we use $HOMEDIR/.kube/config, except on windows where we use $USERPROFILE/.kube/config.

This is the same logic that kubectl uses to determine the kubeconfig, so maybe we can just refer how that detects kubeconfig?

content/kots-cli/download.md
```diff
@@ -5,4 +5,24 @@ title: kots download
...
+### Usage
+```bash
+kubectl kots download [app-slug] [flags]
+```
+* _Replace `[app-slug]` with the "slug" of your KOTS application_ 
+* _Provide `[flags]` according to the table below_ 
+
+| Flag                 | Type | Description |
+|:----------------------|------|-------------|
+| `--dest` | string |        the directory to store the application in (default "/Users/austin") |
```
marccampbell: This defaults to your home directory ($HOME var for Mac and Linux, $USERPROFILE dir on Windows). I think that this doc should not contain your homedir, but an indicator that this default is environment specific.

content/kots-cli/download.md
```diff
@@ -5,4 +5,24 @@ title: kots download
...
+| `-h, --help`   |  |           help for download |
+| `--kubeconfig` | string |  the kubeconfig to use (default "/Users/austin/.kube/config") |
+| `-n, --namespace` | string |    the namespace to download from (default "default") |
+| `--overwrite` |   |       overwrite any local files, if present |
```
marccampbell: this is a `bool` type.

marccampbell: A quick glance, I think all of the missing types in this doc are bool types

content/kots-cli/upload.md
```diff
@@ -5,4 +5,26 @@ title: kots upload
 weight: 90140
 ---
 
-If you have a copy of an application that was created with kots pull or kots download, you can upload it back to the Admin Console using the `kots upload` command.
+Upload Kubernetes manifests from the local filesystem to a cluster, creating a new version of the application that can be deployed. When you have a copy of an application that was created with `kots pull` or `kots download`, you can upload it back to the Admin Console using the `kots upload` command.
```
marccampbell: ~to a cluster~ to the Admin Console?  I wonder if saying that it uploads to the cluster could be interpreted as a way of saying that the application is deployed?  This command just makes it available as a new version in the Admin Console.

content/kots-cli/reset-password.md
```diff
@@ -6,3 +6,21 @@ weight: 90160
 ---
 
 If you’ve deployed an application with the admin console, the `kots reset-password` command will change the bcrypted password hash in the cluster, allowing you to log in again.
+
+### Usage
+```bash
+kubectl kots reset-password [namespace] [flags]
+```
+* _Replace `[namespace]` with the namespace where your KOTS application resides_
```
marccampbell: Specifically, it's not where the application resides, but where the Admin Console is deployed

content/kots-cli/download.md
austinchambers: @marccampbell For all the issues with KOTS command-line flags, these were copied from kots --help, so if we want to change it here, we should probably also change it there. 

As for the missing types, I was thinking that when unspecified, it is not followed by a value. For example, I think usage is '--overwrite' and not '--overwrite true'

content/kots-cli/download.md
austinchambers: What is displayed is what is rendered by "--help". I'll go ahead and add the helpful details about the flag.

content/kots-cli/download.md
austinchambers: I'll go ahead and add the helpful details about the flag. If we change it here, we may also want to change what is displayed for --help.

content/kots-cli/upload.md
austinchambers: Thanks. Will adjust wording.

austinchambers: Note that the wording from "--help" is presently "Upload Kubernetes manifests from the local filesystem to a cluster, creating a new version of the application that can be deployed." So we probably want to update that as well.

content/kots-cli/reset-password.md
austinchambers: Will update wording.

content/kots-cli/download.md
austinchambers: Saying default is $KUBECONFIG. If unset, then $HOME/.kube/config. Windows users generally know that $HOME = $USERPROFILE (saying this messed with the column presentation)

content/kots-cli/download.md
austinchambers: Saying (defaults to $HOME).

content/kots-cli/download.md
austinchambers: I'm inclined to keep this as-is, and update if we update the "--help" command.

PR #111, opened 2020-02-10
fix config file and ConfigOptionData documentation

laverya:


content/reference/v1beta1/config.md
```diff
@@ -141,25 +141,17 @@ A `textarea` can specify a `props` that will map into the HTML element directly.
 ```
 
 ### `file`
-A `file` is a special type of form field that renders an [`<input type="file" />`](https://www.w3schools.com/tags/tag_input.asp) HTML element. It has both a value, which holds the file name, and data, the contents of the file. See the [`ConfigOptionData`](https://help.replicated.com/docs/native/packaging-an-application/template-functions/#configoptiondata) template function for examples on how to use the file contents in your application.
+A `file` is a special type of form field that renders an [`<input type="file" />`](https://www.w3schools.com/tags/tag_input.asp) HTML element. It has both a value, which holds the file name, and data, the contents of the file. See the [`ConfigOptionData`](hhttps://kots.io/reference/template-functions/config-context/#ConfigOptionData) template function for examples on how to use the file contents in your application.
```
dexhorthy: you have double-h here, and how about a relative path anyways?

```suggestion
A `file` is a special type of form field that renders an [`<input type="file" />`](https://www.w3schools.com/tags/tag_input.asp) HTML element. It has both a value, which holds the file name, and data, the contents of the file. See the [`ConfigOptionData`](/reference/template-functions/config-context/#ConfigOptionData) template function for examples on how to use the file contents in your application.
```

PR #113, opened 2020-02-19
Annotations cannot be booleans

emosbaugh:


content/vendor/packaging/optional-resources.md
```diff
@@ -56,7 +56,7 @@ kind: Statefulset
 metadata:
   name: postgresql
   annotations:
-    "kots.io/exclude": repl{{ ConfigOptionEquals "postgres_type" "embedded_postgres" }}
+    "kots.io/exclude": 'repl{{ ConfigOptionEquals "postgres_type" "embedded_postgres" }}'
```
laverya: if you use `{{repl` then it'll keep quotes after going through kustomize
```suggestion
    "kots.io/exclude": '{{repl ConfigOptionEquals "postgres_type" "embedded_postgres" }}'
```

PR #114, opened 2020-02-20
[1.13.1] Ingress

emosbaugh:


content/vendor/packaging/ingress.md
```diff
@@ -0,0 +1,98 @@
...
+### Ingress
+
+```yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: example-application-ingress
+  annotations:
+    "kots.io/exclude": 'repl{{ or (ConfigOptionEquals "enable_ingress" "1" | not) IsKurl }}'
+    "kubernetes.io/ingress.allow-http": 'repl{{ if ConfigOptionEquals "allow_http" "1" }}truerepl{{ else }}falserepl{{ end }}'repl{{ ConfigOption "annotations" | nindent 4 }}
```
austinchambers: Isn't the above GKE-specific? (i.e., ingress.allow-http?) 

Why not: 
```
  annotations:
    nginx.ingress.kubernetes.io/force-ssl-redirect: 'repl{{ (ConfigOptionEquals "allow_http" "1" | not) }}'repl{{ ConfigOption "annotations" | nindent 4 }}
```

content/vendor/packaging/ingress.md
```diff
@@ -0,0 +1,98 @@
...
+              serviceName: nginx
+              servicePort: 80
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: example-application-ingress
+  annotations:
+    "kots.io/exclude": 'repl{{ or (ConfigOptionEquals "enable_ingress" "1" | not) (not IsKurl) }}'
+    "kubernetes.io/ingress.allow-http": 'repl{{ if ConfigOptionEquals "allow_http" "1" }}truerepl{{ else }}falserepl{{ end }}'repl{{ ConfigOption "annotations" | nindent 4 }}
```
austinchambers: Same as above. Also, why are we excluding this section when ingress is enabled?

content/vendor/packaging/ingress.md
```diff
@@ -0,0 +1,98 @@
...
+        - name: hostname
+          type: text
+          title: Hostname
+          description: |
+            Use this field to provide a hostname for your Example Application installation.
+          required: true
+          when: repl{{ ConfigOptionEquals "enable_ingress" "1" }}
+        - name: allow_http
+          type: bool
+          title: Allow HTTP?
```
austinchambers: How about 
```
        - name: allow_http
          title: Allow Unsecured Access through HTTP
          type: bool
          value: "1"
```

austinchambers:
Could also use some steps to set this up. For some reason, it doesn't work for me.

austinchambers:
Maybe this is because kots.io exclude doesn't work? 
image

emosbaugh:
Makes sense. It probably has to be two separate files. Try that. I’ll
address all comments tomorrow

content/vendor/packaging/ingress.md
emosbaugh: i didnt realize that `kubernetes.io/ingress.allow-http` is gke specific. couldnt hurt to add both annotations though. i will add `nginx.ingress.kubernetes.io/force-ssl-redirect`

emosbaugh:
> IsKurl is not defined. This will mean that folks following the guide will get stuck on uploading the license. I also get an error deploying the application because there are situations where having two ingresses of the same name is possible.

I was being optimistic that this would be added before the doc is released. otherwise i could not make a solution that works for both types of clusters.

content/vendor/packaging/ingress.md
emosbaugh: will do

content/vendor/packaging/ingress.md
emosbaugh: we excluded when ingress IS NOT enabled or it IS NOT an embedded cluster

content/vendor/packaging/ingress.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Ingress
+
+```yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: example-application-ingress
+  annotations:
+    "kots.io/exclude": 'repl{{ or (ConfigOptionEquals "enable_ingress" "1" | not) IsKurl }}'
```
emosbaugh: this is excluded when ingress IS NOT enabled or it IS an embedded cluster

austinchambers: Yeah, the reason they were both included was due to that bug where multiple exclusions in a single file don't work.

PR #138, opened 2020-03-12
Ingress IsKurl switched for embedded and existing clusters

emosbaugh:


content/vendor/packaging/ingress.md
```diff
@@ -87,7 +87,7 @@ kind: Ingress
 metadata:
   name: example-application-ingress-embedded
   annotations:
-    kots.io/exclude: '{{repl or (ConfigOptionEquals "enable_ingress" "1" | not) (not IsKurl) }}'
+    kots.io/exclude: '{{repl or (ConfigOptionEquals "enable_ingress" "1" | not) IsKurl }}'
```
MikaelSmith: Was this a mistake? It used to *exclude* the embedded version if `not IsKurl`. Now it will select the non-embedded version if `IsKurl`. I thought what was here before was correct (though the double negative makes it confusing).

emosbaugh: thanks! this is incorrect. reverting.

PR #183, opened 2020-04-28
Adding support for windows; Making the `deps` directory optional

austinchambers:
We lost Windows support by downloading Hugo directory. Adding back, by request. 

Also: 
* Making the /deps/ directory and html proofing optional. This way, you have a choice if you want `make dev` to use official binaries or whatever you have locally. 
* I noticed that the generated SCSS files were added back to the repo for some reason. Removed those again. 

Verified in Windows, Ubuntu, Codeserver, and macOs.


Makefile
```diff
@@ -41,11 +38,15 @@ install:
...
 
 .PHONY: test
-test: deps deps-proofer
+test: 
 	rm -rf public
-	deps/hugo -v -s .
-	htmlproofer --allow-hash-href --check-html --empty-alt-ignore --url-ignore /kots.io/css/ "./public"
\ No newline at end of file
+	# Use hugo dependencies via `make deps`, if available. Otherwise use OS install of Hugo.
+	$(DIR)hugo -v -s .
```
dexhorthy, line 49: what do you think about pulling `$(DIR)hugo` into ` `$(HUGO)` var?

dexhorthy: then it becomes a make var that folks can override to a specific path if they want

Makefile
```diff
@@ -41,11 +38,15 @@ install:
...
+	
+	# Run htmlproofer 3.15.0, if installed. Can be installed with `gem install --user-install html-proofer -v 3.15.0`
+	-htmlproofer --allow-hash-href --check-html --empty-alt-ignore --url-ignore /kots.io/css/ "./public"
```
dexhorthy, line 52: excuse me if this is me being bad at windows -- but did this leading hyphen get added intentionally?

austinchambers: It was intentional. It's a make comment. The hyphen causes it to not cause the make test to exhibit a failure.

Makefile
austinchambers: Good idea.

austinchambers: OK, updated accordingly.

Makefile
dexhorthy: gotcha

PR #186, opened 2020-05-01
Add a contributing and style guide

marccampbell:
These are carried over and edited a touch from the Replicated Help Center style guides.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,94 @@
+## Contributing Guidelines
+When contributing to the KOTS documentation, please consider the following writing styles:
+
+- ### Avoid acronyms. If using one, make sure to define it on first use.
+
+- ### Favor writing in the [active voice](http://writing.wisc.edu/Handbook/CCS_activevoice.html).
+**Yes:**  The next section includes some basic information about our app release including the app name.
+
+**No:**  In the next section some basic information is included about our app release including the app name.
+
+- ### Favor writing in the [indicative mood](https://grammarist.com/grammar/english-moods/).
+**Yes:**  At runtime, Replicated will rewrite the image tag to pull from `registry.replicated.com`.
+
+**No:**  At runtime, Replicated would rewrite the image tag to pull from `registry.replicated.com`.
+
+- ### Avoid Second Person Pronouns
+**Yes:** A license value can be used to show and hide fields on the settings page.
+
+**No:** You can use your license to allow customer access to hidden settings.
+
+- ### Avoid First Person Pronouns
+**Yes:** The Go playground can be used to validate the regular expression.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,94 @@
...
+
+**No:** A Kots application can include...
+**No:** A kots application can include...
+
+- ### The product and the company are named Replicated (with a capital R).
+**Yes:** How to install Replicated.
+
+**No:** How to install replicated.
+
+- ### The kotsadm Admin Console is referred to as the Admin Console
```
dexhorthy, line 83: This is the only one I have issue with -- I think kotsadm is fairly succint.

What about kURL? Should we call these "kURL clusters"? "kURL Embedded Clusters"? "KOTS embedded cluster"?

austinchambers: My vote is "embedded cluster," since it's the closest to what we communicate in Vendor. You have two installation options with KOTS: "embedded cluster" and "existing cluster." The fact that kURL is used to help us achieve the former is presently a technical detail that we can choose to communicate. 

As another example, I see "Support Bundling" as a feature of KOTS. We can choose to communicate that it's backed by the troubleshoot project, but that's also a technical detail (we haven't been calling it "troubleshoot.sh Support Bundling").

CONTRIBUTING.md
```diff
@@ -0,0 +1,94 @@
...
+
+**No:**  ```For more information, read the release notes <a href="/link">here</a>.```
+
+- ### Only use Markdown backticks on code or command literals.
+The backtick symbol shouldn't be used for emphasis.
+
+- ### Code snippets in bash should not include sudo.
+**Yes:**  ```docker logs -f replicated```
+
+**No:** ```sudo docker logs -f replicated```
```
austinchambers, line 48: I'd vote for this one to be removed. I actually prefer the latter, since it communicates that higher privileges is required for the command and that it has the highest change of working as-is. For example, we shouldn't remove the sudo from `curl ... | sudo bash` 

Or maybe we can revise with `don't use stuff in example commands if it's not needed`, if that really needs to be said.

CONTRIBUTING.md
dexhorthy: @austinchambers  do you want to suggest a change here? or is this good to go?

CONTRIBUTING.md
dexhorthy: @marccampbell what do you think? I'd like to get this merged in soon. 

@austinchambers while your argument isn't totally off and I'm interested to hear why we originally settled on the `no sudo in commands` convention, I think we've had this convention in place on all our docs sites for years now, and changing it should take place in a separate proposal.

PR #194, opened 2020-05-04
Enabling pdf content generation when visiting /all-content/

austinchambers:
Was needed to export to PDF. Simply go to https://deploy-preview-194--kotsio.netlify.app/all-content to see it.


dexhorthy:
I am on board with this 👍

PR #201, opened 2020-05-11
update docs with correct tar.gz semantics

dexhorthy:


content/kotsadm/installing/installing-embedded-cluster.md
```diff
@@ -21,17 +21,20 @@ curl -sSL https://kurl.sh/supergoodtool | sudo bash
...
-tar xzvf supergoodtool.tar.gz
+curl -SL -o supergoodtool.tar https://kurl.sh/bundle/supergoodtool.tar.gz
+tar xvf supergoodtool.tar
 cat install.sh | sudo bash -s airgap
 ```
 
 Note that the airgapped installer is not the same as an [airgapped application package](/kotsadm/installing/airgap-packages/).
 A KOTS application may be installed in airgap mode on clusters installed online and vice versa.
 
+kURL currently uses `.tar.gz` extension for a `.tar` file, hence the `-o *.tar`.
```
aj-jester, line 35: 😹 👍

PR #207, opened 2020-05-14
Helm V3

emosbaugh:


content/reference/v1beta1/helmchart.md
```diff
@@ -25,6 +25,9 @@ spec:
 
   exclude: "repl{{ ConfigOptionEquals `include_chart` `include_chart_no`}}"
 
+  # helmVersion identifies the Helm Version used to render the Chart. Default is v2.
+  helmVersion: v3
```
austinchambers, line 29: Since v3 is beta, I'm thinking it's best to show this `helmVersion` field, but use `v2` in the example.

content/kots-cli/pull.md
```diff
@@ -24,6 +24,7 @@ This command supports all [global flags](/kots-cli/global-flags/) and also:
 | `--exclude-admin-console` |  bool  |  set to true to exclude the admin console _(only valid when `[upstream-uri]` points to a replicated app)_ |
 | `--exclude-kots-kinds` |  bool  | set to true to exclude rendering kots custom objects to the base directory _(default `true`)_ |
 | `-h, --help`  |          |  help for pull |
+| `--helm-version` | string | the Helm version with which to render the Helm Chart _(default `"v2"`)_ |
```
austinchambers, line 27: I'm thinking we should probably indicate that it's beta here, and in the CLI.

content/reference/v1beta1/helmchart.md
emosbaugh: but that would be default... why not just omit it then?

content/kots-cli/pull.md
emosbaugh: this is just the cli output

emosbaugh:
Can we get the beta tag to show for helm v3?
- idk. who is responsible for the ui? seems out of scope of this pr

content/reference/v1beta1/helmchart.md
austinchambers: By including "helmVersion: v2", you're documenting how the setting works. If you omit it, it's less explicit (and therefore less useful).

austinchambers:
> Can we get the beta tag to show for helm v3?
> 
> * idk. who is responsible for the ui? seems out of scope of this pr

Ah, I just noticed that it was there already. You've added it.

content/reference/v1beta1/helmchart.md
```diff
@@ -25,6 +25,9 @@ spec:
 
   exclude: "repl{{ ConfigOptionEquals `include_chart` `include_chart_no`}}"
 
+  # helmVersion identifies the Helm Version used to render the Chart. Default is v2.
+  helmVersion: v2
```
dexhorthy, line 29: it says `apiVersion` below -- which one is correct?

emosbaugh: good catch. fixed

PR #229, opened 2020-07-09
add note about hash not being special, namespace example

dexhorthy:


content/vendor/guides/production-ready.md
```diff
@@ -53,10 +53,40 @@ If you expect to also install stateful services into existing clusters, you'll l
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: spline-reticulator
+  namespace: graphviz-pro
+spec: 
+```
+
+```yaml
+# bad, configurable
```
austinchambers, line 81: Saying this is bad seems to contradict the earlier opinion we state: “Letting the end user manage namespaces will be the easiest way to reduce friction...”
Maybe there’s a better way to word this? “Less good, unless configured through kots?”

dexhorthy: This is a good question. Let me take another pass and clarify what I mean by `Letting the end user manage namespaces`

aj-jester: May be we can say "let the end user manage namespace and in your YAML add something like `namespace: repl{{ Namespace }}`' 🤷

PR #231, opened 2020-07-13
1.17.0 release notes

salahalsaleh:


content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,42 @@
+---
+date: 2020-07-13T08:00:00-07:00
+linktitle: "1.17.0"
+title: "1.17.0"
+kubernetes: "1.16, 1.17, and 1.18"
+---
+
+{{<features>}}
+* GitOps - Added Github Enterprise support.
+* Snapshots - Added a "Internal Storage (Default)" option to the providers list for KURL embedded clusters. Which will switch back to the default configuration that velero was set up with when the app was installed.
```
salahalsaleh, line 10: @austinchambers there's probably a better wording for this. any ideas?

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
+---
+date: 2020-07-13T08:00:00-07:00
+linktitle: "1.17.0"
+title: "1.17.0"
+kubernetes: "1.16, 1.17, and 1.18"
+---
+
+{{<features>}}
+* GitOps - Added Github Enterprise support.
```
austinchambers, line 9: This is a bug, not a new feature. That being said, I understand why we want to have this listed as a "new feature." How about we treat the bug as "We incorrectly allowed GitHub Enterprise as an option when it wasn't supported yet." Something like that. 

```suggestion
* Admin Console now supports GitHub Enterprise as a GitOps repository. 
```

content/release-notes/1.17.0.md
austinchambers:
```suggestion
* Ability to change to a different snapshot storage provider using the UI on the Snapshot Settings page. 
* "Internal Storage" snapshot provider option has been added to the snapshot providers list for kURL embedded clusters. This allows the user to easily return to the default snapshot configuration when another provider (such as AWS S3 or Google Storage) has been configured. 
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Snapshots - Added a "Internal Storage (Default)" option to the providers list for KURL embedded clusters. Which will switch back to the default configuration that velero was set up with when the app was installed.
+* Added `when` field support for config groups.
```
austinchambers, line 11:
```suggestion
* Added `when` field support for config groups [https://kots.io/reference/v1beta1/config/#when](https://kots.io/reference/v1beta1/config/#when)
```

austinchambers: Verify the above syntax renders properly. I think I got it right.

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+{{</features>}}
+
+{{<changes>}}
+* Kotsadm - New Support bundles will be shown at the top of the list instead of the bottom.
```
austinchambers, line 15:
```suggestion
* The most recent support bundles on the Admin Console troubleshoot page are now shown first. 
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Kotsadm - Showing warnings on the Version History page in the Admin Console when installing an application with invalid YAML.
```
austinchambers, line 16:
```suggestion
* Helpful warnings now appear on the Version History page if the user attempts to deploy a version of the application with invalid YAML. 
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Kotsadm - Added a page title for the 'View files' and 'Config' tabs.
```
austinchambers, line 17: This is kind of small. I'd remove. 
```suggestion
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Removed the KOTS metadata that was added to all application manifests. Which included the local application and the version sequence injected as annotations to all deployed manifests.
```
austinchambers, line 18: Why did we do this?

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Show the airgap upload progress percentage more granularly.
```
austinchambers, line 19:
```suggestion
* Improved the granularity of the airgap upload progress indicator.
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Removed the "Re-analyze" button from the troubleshoot support bundle analysis page.
```
austinchambers, line 20: Why was this removed?

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Auto deploy the app's first version when installing the app and no preflight checks are available or preflight checks pass.
```
austinchambers, line 21: I thought we already did this? Or is this just for existing clusters? What was the experience like when we didn't do this and no preflights were configured?

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Update to import k8s 1.18 libraries.
```
austinchambers, line 22: 🤷  Support for later versions of k8s is one thing, but I don't care about which libraries are imported. 
```suggestion
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Show a descriptive error message when `kubectl kots download` is called with an app slug that does not exist.
```
austinchambers, line 23:
```suggestion
* Improved error message when `kubectl kots download` is called with an app slug that does not exist.
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+{{</changes>}}
+
+{{<fixes>}}
+* Fixed a bug in KOTS CLI where using `--deploy` flag when checking for updates does not deploy the latest release.
```
austinchambers, line 27:
```suggestion
* Fixed a bug in KOTS CLI where the `--deploy` flag when checking for [upstream](https://kots.io/kots-cli/upstream/) updates fails to deploy the latest release.
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed a bug in KOTS CLI where automated installs fail to deploy if no preflight checks are available.
```
austinchambers, line 28: Which CLI command is this for? Guessing 'install'? 

```suggestion
* Fixed a bug in KOTS CLI [install command](https://kots.io/kots-cli/install/) where automated installs fail to deploy if no preflight checks are configured for the application. 
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed a bug in KOTS CLI where checking for updates shows `There are no application updates available` even when updates are actually available.
```
austinchambers, line 29: Which command was this for? Is it `kots version`? Is it `upstream`?

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed an issue where license expiration isn't checked until after fetching application.
+* Fixed a bug where preflight checks were not triggered for airgap or automated installs.
```
austinchambers, line 31: Does this always happen, or just in a specific situation? If so, what situation? If not, I'd use stronger wording to convey this is seriously broken (and probably put this at the top of the bug fixes list). Also, what do you mean by `automated installs`? Is this an install via `kots install?`

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed a bug where Snapshots scheduled `@daily` run just on day 1 of the month instead of every day.
+* Fixed an issue in Kotsadm troubleshoot "View Files" tab that shows full path instead of just subdirectory.
```
austinchambers, line 33:
```suggestion
* Fixed an issue on the "View Files" tab in the Admin Console that shows full path instead of just subdirectory.
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed an issue where the minio image included a binary with CAP_CHOWN set.
```
austinchambers, line 34: Ok... but what's the issue? What happens when this binary is included? If there's no user impact, I'd probably remove this line.

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed button styles on TLS page on KOTS cluster installs.
```
austinchambers, line 35: This was a visual glitch. 

```suggestion
* Fixed visual glitches on the buttons of the [TLS certificate page](https://kurl.sh/docs/install-with-kurl/setup-tls-certs) that appears during embedded cluster installation. 
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed snapshot hooks status.
```
austinchambers, line 36: Not sure what this is. Need more detail, I think. How did this bug manifest to the end user?

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed backup size on snapshots list.
```
austinchambers, line 37: Is this when the backup size incorrectly says `0 bytes` or `undefined` or something like that? 

If so, I'd explicitly state that.

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed Velero checks on Snapshot settings.
```
austinchambers, line 38: Not sure what this is about. Is this about the checks that fail because we require elevated permissions? What's the user experience here?

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed a bug where clicking `Save changes` in the "Registry settings" tab without changing anything breaks the registry.
```
austinchambers, line 39: I'd put this near the top of the bug list. Corruption is high severity. 

```suggestion
* Fixed a bug where clicking `Save changes` in the "Registry settings" tab without changing anything permanently breaks the registry connection to the Admin Console. Repairing this connection then requires manual intervention outside the Admin Console. 
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed an issue where the diff viewer in Kotsadm does not scroll to the end of the file.
+* Fixed an issue where pushing images during an airgap installation just hangs.
```
austinchambers, line 41: I believe it just appears to hang. Adding my understanding below. Correct if wrong. 

```suggestion
* Fixed an issue where pushing images during an airgap installation hangs when the progress bar is nearly complete. Attempting to reload the page will either fix the problem, or prompt another upload that will work the second time around. 
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed a bug where automatic update checks result in duplicate releases.
```
austinchambers, line 42:
```suggestion
* Fixed a bug where automatic update checks result in duplicate releases on the Version History page.
```

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,44 @@
...
+* Fixed a bug where re-uploading an airgap bundle does not work if previous upload fails.
```
austinchambers, line 43: This always happens? When the user is in this situation, how can they fix it?

content/release-notes/1.17.0.md
salahalsaleh: It was force restarting pods for customers, so we decided to get rid of them for the time being and come with a better solution in the future. Also, these annotations are not used at all yet.

content/release-notes/1.17.0.md
salahalsaleh: It was not working and it served no significant purpose.

content/release-notes/1.17.0.md
salahalsaleh: Hmmm, you're right, we should probably move this to the "Fixes" section.
This was not working for:
1- Automated installs using the CLI.
2- For apps with no preflight checks.
3- If "Continue" is not clicked in the preflights page.

content/release-notes/1.17.0.md
salahalsaleh: `kots upstream upgrade`

content/release-notes/1.17.0.md
salahalsaleh: This always happens. And yes, by automated installs i mean using `kots install` and providing a license and config values, which should bypass the config and preflights screen, in which case, the api was not triggering the preflights.

content/release-notes/1.17.0.md
salahalsaleh: Hmmmm, i guess not, it only happens if the previous upload failed while pushing the app archive to s3/docker.

content/release-notes/1.17.0.md
austinchambers: I'd then put this under the bug section, and say: 
```
Fixed an issue where pods were force-restarted because of KOTS metadata in application manifests. Fixed by ensuring that the system no longer adds KOTS metadata to these manifests. 
```

content/release-notes/1.17.0.md
austinchambers: Ok

content/release-notes/1.17.0.md
austinchambers: Sounds like it should be in the "Fixes" section. 

Still don't quite understand what happened. But if you pair that up with the 3 items above, that sounds good to me.

content/release-notes/1.17.0.md
austinchambers: Then I'd add this release note right below the other `upstream` release note.
```suggestion
* Fixed a bug in [upstream](https://kots.io/kots-cli/upstream/) KOTS CLI where checking for updates shows `There are no application updates available` even when updates are actually available.
```

content/release-notes/1.17.0.md
marccampbell: I think this should stay as a change. Removing this could possibly have adverse effects on an app if they were using undocumented behavior. Specifically this was removed ONLY for gitops enabled clusters (which should be called out)

content/release-notes/1.17.0.md
marccampbell: @austinchambers but kots can be used as a library, and i'd rather include developer release notes than not include them. This is important for someone who might be using kots in a way that we don't document specifically.

content/release-notes/1.17.0.md
marccampbell: This changes the security requirements for KOTS. This should be reworded and kept.

How about:

```Removed the CAP_CHOWN capability from the minio docker file. When deploying to existing clusters, this SELinux capability will no longer be required.```

File: content/release-notes/1.17.0.md (hunk 7 above)
Comment (austinchambers) [reply content/release-notes/1.17.0.md:20]: In that case, I'd advocate it being both a change and a bug. I'd also call out GitOps enabled clusters. 

```suggestion
* Removed the KOTS metadata that was added to all application manifests. For GitOps enabled clusters, this metadata could result in situations where pods were incorrectly force-restarted. 
```

content/release-notes/1.17.0.md
austinchambers: I suppose developer release notes are possibly best placed in the repository, and not kots.io. It confuses our audiences. So I still think it's best removed. 

Also curious about how you use KOTS as a library?

content/release-notes/1.17.0.md
austinchambers: Sounds good to me: 
```suggestion
* Removed the CAP_CHOWN capability from the minio docker file. When deploying to existing clusters, this SELinux capability will no longer be required.
```

content/release-notes/1.17.0.md
marccampbell: This is not correct.  We only added it to GitOps clusters.

content/release-notes/1.17.0.md
marccampbell: It's a go library (actually kotsadm uses it as a go library, not an executable).

Let's reword this to:

* Update kubernetes dependencies to 1.18

content/release-notes/1.17.0.md
austinchambers:
```suggestion
* Fixed a bug where preflight checks fail to trigger for UI-based airgap installations, or for automated installations where license and configuration has been provided to `kots install` via the CLI. 
```

content/release-notes/1.17.0.md
austinchambers: Rewording accordingly
```suggestion
* Update to import k8s 1.18 libraries.
```

marccampbell: I think we broke the github suggested change here.

content/release-notes/1.17.0.md
austinchambers: Ok, tried to update accordingly.

content/release-notes/1.17.0.md
```diff
@@ -0,0 +1,46 @@
...
+
+{{<changes>}}
+* Velero add-on in kURL has been updated to match the official Velero configuration.  This is a breaking change that requires the velero namespace to be deleted before performing the next update with kURL.
+* The most recent support bundles on the Admin Console troubleshoot page are now shown first. 
+* Helpful warnings now appear on the Version History page if the user attempts to deploy a version of the application with invalid YAML. 
+* Removed the KOTS metadata that was added to all application manifests for GitOps enabled clusters. This metadata could result in situations where pods were incorrectly force-restarted. 
+* Improved the granularity of the airgap upload progress indicator.
+* Removed the "Re-analyze" button from the troubleshoot support bundle analysis page.
+* Auto deploy the app's first version when installing the app and no preflight checks are available or preflight checks pass.
+* Update to import k8s 1.18 libraries.
```
marccampbell, line 24: Update kubernetes dependencies to 1.18

PR #242, opened 2020-07-23
Hidden vs. Readonly Properties

aj-jester:
- Document how `RandomString` behaves based on `hidden` and `readonly` fields.
-  `help_text` docs were missing and clarified usage of `description`.
    - Also added a note that both support Markdown syntax.
- Updated KOTS-K8s compatibility matrix.

### DIFF
--- a/content/reference/template-functions/static-context.md
+++ b/content/reference/template-functions/static-context.md
@@ -150,7 +150,6 @@ Returns a random string with the desired length and charset.
 Provided charsets must be Perl formatted and match individual characters.
 If no charset is provided, `[_A-Za-z0-9]` will be used.
 
-Each time that this function is called, it will return a different value.
 ```yaml
 '{{repl RandomString 64}}'
 ```
@@ -158,6 +157,13 @@ Or for a total of 64 `a`s and `b`s:
 ```yaml
 '{{repl RandomString 64 "[ab]" }}'
 ```
+
+Each time that this function is called, the behavior changes based on the [hidden](/reference/v1beta1/config/#hidden) and [readonly](/reference/v1beta1/config/#readonly) properties.
+
+- To generate `RandomString` value that is **persistent** between Config changes, use it in conjuction with `hidden` property set to `true`. The `value` is not shown in HTML, hence it cannot be modified.
+- To generate `RandomString` value that is **ephemeral** between Config changes, use it in conjuction with `readonly` property set to `true`. The `value` is shown in HTML but the it cannot be modified.
+- If `hidden` and `readonly` are not set or set to `false`, the `value` is **persistent** between Config changes but it can also be modified in HTML.
+- If both `hidden` and `readonly` are set to `true`, the `value` is not shown in HTML but it is **ephemeral** between Config changes and not modifiable.
 
 ## Add
 ```go
--- a/content/reference/v1beta1/config.md
+++ b/content/reference/v1beta1/config.md
@@ -256,6 +256,6 @@ This is similar to `description` but for `items`. This property can show a helpf
       items:
       - name: http_enabled
         title: HTTP Enabled
-        help_text: When enabled we will listen to http
+        help_text: Check to enable the HTTP listener
         type: bool
 ```
--- a/content/reference/v1beta1/config.md
+++ b/content/reference/v1beta1/config.md
@@ -250,7 +250,7 @@ Items can be affixed left or right. These items will appear in the admin console
 ```
 
 ### `help_text`
-This is similar to `description` but for `items`. This property can show a helpful message below `title`. **[Markdown](https://guides.github.com/features/mastering-markdown/) syntax is supported.**
+This is similar to `description` but for `items`. This property can show a helpful message below `title`. [Markdown](https://guides.github.com/features/mastering-markdown/) syntax is supported.
 ```yaml
     - name: toggles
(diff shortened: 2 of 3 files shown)


content/reference/v1beta1/config.md
```diff
@@ -235,9 +239,23 @@ Items can be readonly.
           readonly: true
 ```
```
emosbaugh, line 241: My understanding is it’s more than just random string

aj-jester: @emosbaugh Oh! So does that mean _any_ function that generate "things" (RandomString being just one of them) will regenerate based on `hidden` vs `readonly`?

content/reference/template-functions/static-context.md
```diff
@@ -159,6 +158,13 @@ Or for a total of 64 `a`s and `b`s:
 '{{repl RandomString 64 "[ab]" }}'
 ```
 
+Each time that this function is called, the behavior changes based on the [hidden](/reference/v1beta1/config/#hidden) and [readonly](/reference/v1beta1/config/#readonly) properties.
+
+- To generate `RandomString` value that is **persistent** between Config changes, use it in conjuction with `hidden` property set to `true`. The `value` is not shown in HTML, hence it cannot be modified.
```
dexhorthy, line 163:
```suggestion
- To generate a `RandomString` value that is **persistent** between Config changes, use it in conjunction with `hidden` property set to `true`. The `value` is not shown in HTML, hence it cannot be modified.
```

content/reference/template-functions/static-context.md
```diff
@@ -159,6 +158,13 @@ Or for a total of 64 `a`s and `b`s:
...
+- To generate `RandomString` value that is **ephemeral** between Config changes, use it in conjuction with `readonly` property set to `true`. The `value` is shown in HTML but the it cannot be modified.
```
dexhorthy, line 164:
```suggestion
- To generate a `RandomString` value that is **ephemeral** between Config changes, use it in conjunction with `readonly` property set to `true`. The `value` is shown in HTML but the it cannot be modified.
```

content/reference/v1beta1/config.md
dexhorthy: I think that's right @aj-jester but I'll let @emosbaugh confirm

content/reference/template-functions/static-context.md
```diff
@@ -159,6 +158,13 @@ Or for a total of 64 `a`s and `b`s:
 '{{repl RandomString 64 "[ab]" }}'
 ```
 
+Each time that this function is called, the behavior changes based on the [hidden](/reference/v1beta1/config/#hidden) and [readonly](/reference/v1beta1/config/#readonly) properties.
```
emosbaugh, line 161: does hidden matter? i dont know

content/reference/v1beta1/config.md
```diff
@@ -225,6 +225,10 @@ Items can be hidden. They will not be visible if hidden.
           value: "{{repl RandomString 40}}"
 ```
 
+When used in conjunction with [RandomString](/reference/template-functions/static-context/#randomstring)
```
emosbaugh, line 228: its my experience that readonly affect this. does hidden?

content/reference/v1beta1/config.md
```diff
@@ -235,9 +239,23 @@ Items can be readonly.
           readonly: true
 ```
 
+When used in conjunction with [RandomString](/reference/template-functions/static-context/#randomstring)
```
emosbaugh, line 242: its more than just random string. i think we should generalize this

content/reference/template-functions/static-context.md
aj-jester: It does not but generally it is used with `hidden`.

content/reference/v1beta1/config.md
aj-jester: Updated to make it more generic.

content/reference/v1beta1/config.md
aj-jester: It doesn't but it generally is used with `hidden` If its not a hidden field they can edit it.

content/reference/v1beta1/config.md
aj-jester: Updated the doc to be more generic.

content/reference/template-functions/static-context.md
dexhorthy: @aj-jester I noticed this was resolved but the change was not applied -- can we fix the typos here?

content/reference/v1beta1/config.md
```diff
@@ -235,9 +239,23 @@ Items can be readonly.
...
 ```
+
+### `help_text`
+This is similar to `description` but for `items`. This property can show a helpful message below `title`. [Markdown](https://guides.github.com/features/mastering-markdown/) syntax is supported.
+```yaml
+    - name: toggles
+      items:
+      - name: http_enabled
+        title: HTTP Enabled
+        help_text: When enabled we will listen to http
```
dexhorthy, line 259: kind of unrelated but I'm not totally a fan of this kind of language in a config screen, and I don't think we should encourage vendors to do this in their demo apps. Who is `we` in this case?


how about

```suggestion
        help_text: Check to enable the HTTP listener
```

PR #249, opened 2020-07-29
New doc for adding a custom application icon

GraysonNull:


markpundsack:
@GraysonNull Looks like this new page shows up as the first content under Admin Console Customization. Is that a conscious choice? Is it one of the first things a vendor would configure?

There doesn't appear to be a page that describes the Application generically. If there were, maybe this content would be better as a section in that page. https://kots.io/vendor/config/dashboard-buttons/#provide-a-kots-application-spec seems the closest in content but is under a Dashboard buttons page, for some reason.

GraysonNull:
@markpundsack  Yeah I saw that page but thought that even though it's somewhat related, that page being all about dashboard buttons might not be the best place for this. Maybe that page should be edited to be more generically about dashboard customization? Then I think the icon docs make sense on that page.

GraysonNull:
Alternatively @markpundsack I recently found this https://kots.io/reference/v1beta1/application/#icon

Seems like maybe the best candidate for where to put these docs

markpundsack:
@GraysonNull Funny, yes, that's probably a better location. But then again, it's buried, and content there is very light, and the content you have wouldn't "fit".

markpundsack:
@GraysonNull Can you move the page to the second-last one in the section?

GraysonNull:
Ok, yeah let me figure out how to reorder those items

GraysonNull:
@markpundsack updated to be second-last in the sidebar

PR #253, opened 2020-08-07
Ha guide

cremerfc:


austinchambers:
Looks like there's an issue with internal anchors: 
```
     <a href="#Verify-the-Deployment">Verify the Deployment</a>
- ./public/vendor/guides/ha-guide/index.html
  *  linking to internal hash #Verify-the-Deployment that does not exist (line 0)
     <a href="#Verify-the-Deployment">Verify the Deployment</a>
  *  linking to internal hash #Verify-the-Deployment that does not exist (line 0)
     <a href="#Verify-the-Deployment">Verify the Deployment</a>
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,410 @@
+---
+date: "2020-02-20T00:00:00Z"
+lastmod: "2020-02-20T00:00:00Z"
+title: "Deploy an HA Cluster"
+weight: "1005"
+---
+
+In this guide, we'll walk through the steps neede to take enable Kubernete's high availability capabilities with a cluster installed by Kurl and managed by Kots. This only applies to 'embedded' cluster installs using Kurl.
```
dexhorthy, line 8: Fixing some typos and splitting to two lines, we prefer "one line per sentence" (I admit we don't follow this rule well though)
```suggestion
In this guide, we'll walk through the steps needed to take enable Kubernetes's high availability capabilities with a cluster installed by kURL and managed by KOTS. 
This only applies to 'embedded' cluster installs using Kurl.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+The diagram shows 4 Virtual Machines (VMs) running on Google Cloud Platform (GCP), with one of them used solely as a load balancer. Note that there aren't any requirements to use GCP. You can follow along with VMs provisioned on a different cloud provider or on premise.
+
+## Overview
+
+The guide is broken into two parts. The [first part](#part-i---setting-up-the-cluster) will walk you through how to stand up the cluster, while the [second part](#part-ii-testing-scenarions--troubleshooting) will go through a couple of testing scenarios to validate the cluster's resiliency and some troubleshooting tips.
+
+There are a few things to keep in mind about this guide:
+
+- All VMs are going to be created in Google Cloud Platform (GCP) but nothing in this guide is dependent on GCP specific services. 
+- The examples in this guide used ephemeral resources.
```
dexhorthy, line 26: What exactly do you mean by ephemeral resources? Servers? IPs? everything?

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+- The sample application has a database component as described [here](#sample-application), which will help us validate data retention in of our testing scenarios.
+
+## Prerequisites
+
+- This exercise will need 4 VMs that will need to communicate with each other. Please refer to the [Procuring the (virtual) hardware](#provisioning-the-virtual-hardware) section for guidance on system requirements for these. 
+
+- In the example commands and screenshots, we are using a sample application called 'AppDirect'. Please refer to the [Sample Application](#sample-application) section for more details if you wish to follow along and use it.
+
+- This is an advanced topic, so this guide will assume you've already completed one of the [Getting Started Guides](/vendor/guides/#getting-started), and have already a level of familiarity with the Replicated tool set.
+
```
dexhorthy, line 37: This is great, love the organization/presentation so far

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+The application consists of two components:
+
+- Python Flask App Deployment - This flask application contains 'routes' to help test connecting, reading and writing to a database. 
+- Postgres StatefulSet - This is the database the for the flask application.
+
+A full description of the application is available in the repository.
+
+### Configuring Ekco
+
+If you want to follow this example, but with your own application, you may need to modify the Kubernetes Installer to use [ekco](https://kurl.sh/docs/add-ons/ekco). Ekco is basically a helper service that helps 'move' scheduled pods from a node that is down to ones that are up.
```
dexhorthy, line 52: I think there may be a more technically succinct way to explain what ekco does, but maybe directing folks to the ekco docs is enough. What do you think of this rephrase?

 
```suggestion
If you want to follow this example, but with your own application, you may need to modify the Kubernetes Installer to use [ekco](https://kurl.sh/docs/add-ons/ekco). For the purposes of this guide, EKCO can be thought of as a helper service that helps 'move' scheduled pods from a node that is down to ones that are up.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+## Part I - Setting Up the Cluster
+
+To stand up the cluster, we will be following the these steps:
+
+- [Procuring the (virtual) hardware](#provisioning-the-virtual-hardware)
+- [Setting up and configuring a Load Balancer](#setting-up--configuring-a-load-balancer)
+- [Run the Kots installer on the first node and deploy the application](#run-the-kots-installer--deploy-application)
+- [Verify the Deployment](#verify-the-deployment)
+- [Add Remaining Nodes to the Cluster](#adding-remaining-nodes-to-cluster)
+- [Verify Pods Are Running on All Nodes](#verify-pods-are-running-on-all-nodes)
```
dexhorthy, line 73: I like this mini-table-of-contents for this section

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+In your Kubernetes installer, add the following under spec:
+
+```code
+  ekco:
+    version: latest
+    nodeUnreachableToleration: 1m
+```
+
+Then make sure to promote the installer to same channel(s) that you are promoting your app.
```
dexhorthy, line 62: The grammar on this one feels a little iffy - do you have any thoughts on a potential re-word?

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+```shell
+gcloud compute instances create app-direct-node-01 --boot-disk-size=200GB --labels=app=app-direct --image-project ubuntu-os-cloud --image-family ubuntu-1804-lts --machine-type n1-standard-4
+```
+To create the remaining nodes, simply run the same command but increment the node number (i.e., "app-direct-node-02").
+
+Also note the VMs public and private ip addresses as each VM is provisioned. We'll use the internal ip addresses in the HA Proxy to configure traffic accordingly.
+
+### Setting Up & Configuring a Load Balancer
+
+The main purpose of the load balancer is to direct traffic to the proper node. All end-user interactions (i.e., access the web UI of the deployed application) should go through the load balancer.
```
dexhorthy, line 105: Is "proper" the right word here? Aren't all the nodes technically able to receive traffic? This seems to imply that the load balancer captures logic about what services are running where.

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+We will:
+
+- Create the config file in our home folder using our favorite editor.
+- Use SCP to copy the file in the VM running HA Proxy.
+- Install Ha Proxy.
+- Copy the config file to the proper location.
+- Restart HA Proxy. 
+
+Note that another way of accomplishing this is by conneting to the instance via SSH and creating the file and editing from the command line inside the VM.
```
dexhorthy, line 119:
```suggestion
Note that another way of accomplishing this is by connecting to the instance via SSH and creating the file and editing from the command line inside the VM.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+To create the config file, use the editor of your choice and call it 'haproxy.cfg'. For the purposes of this exercise, we'll create the file in our home folder. Below is an example config file for the sample application. Keep in mind the following when creating your file:
+
+- IP addresses below will not work as the VMs were destroyed a long time ago. Replace these with the internal IP addresses of your VMs.
```
dexhorthy, line 123: What do you think about dropping the first sentence here?

```suggestion
- Replace any IP addresses with the internal IP addresses of your VMs. 
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+- Copy the config file to the proper location.
+- Restart HA Proxy. 
+
+Note that another way of accomplishing this is by conneting to the instance via SSH and creating the file and editing from the command line inside the VM.
+
+To create the config file, use the editor of your choice and call it 'haproxy.cfg'. For the purposes of this exercise, we'll create the file in our home folder. Below is an example config file for the sample application. Keep in mind the following when creating your file:
+
+- IP addresses below will not work as the VMs were destroyed a long time ago. Replace these with the internal IP addresses of your VMs. 
+- If you used a different naming convention for your instance names, you will need to replace 'app-direct-node-0(123)' with your instance names.
+- There are two entries specific to the Application, one for each component (app-direct-frontend/backend & postgres frontend/backend). If you are using a different application, adjust based on the endpoints you want to access through the load balancer. The remaining entries are for the KOTs UI and API and should be left as-is other than the instance names and ip addresses
```
dexhorthy, line 125:
```suggestion
- There are two entries specific to the Application, one for each component (app-direct-frontend/backend & postgres frontend/backend). If you are using a different application, adjust based on the endpoints you want to access through the load balancer. The remaining entries are for the KOTs UI and API and should be left as-is other than the instance names and ip addresses.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+For more details on the structure of the config file, please refer to the HA Proxy [documentation](https://www.haproxy.com/blog/the-four-essential-sections-of-an-haproxy-configuration/)
+
+To copy the config file into the VM, we are going to use scp. The command below will copy it to your home folder on the VM.
+
+```shell
+gcloud compute scp haproxy.cfg app-direct-ha-proxy:~/
+```
+
+Now we install Ha Proxy using apt.
```
dexhorthy, line 199: I noticed this is sometimes `HA Proxy` and sometimes `Ha Proxy` -- what should we standardize on here?

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+![Channels](/images/guides/kots/ha-cluster-portal-channels.png)
+
+However, in order to enable High Availability we are going to add the '-s ha' [advanced option](https://kurl.sh/docs/install-with-kurl/advanced-options)
+
+```shell
+curl -sSL https://k8s.kurl.sh/appdirect-unstable | sudo bash -s ha
+```
+
+When you include the '-s ha' option, you will be promoted right away for the load balancer address. As shown in the figure below, we are providing the 'internal' IP address. This address is used by the Kubernetes services in the nodes to talk to each other. Using a public IP addresses addds extra layers that are not needed.
```
dexhorthy, line 238:
```suggestion
When you include the '-s ha' option, you will be prompted right away for the load balancer address. As shown in the figure below, we are providing the 'internal' IP address. This address is used by the Kubernetes services in the nodes to talk to each other. Using a public IP address would add extra layers that are not needed. 
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+However, in order to enable High Availability we are going to add the '-s ha' [advanced option](https://kurl.sh/docs/install-with-kurl/advanced-options)
+
+```shell
+curl -sSL https://k8s.kurl.sh/appdirect-unstable | sudo bash -s ha
+```
+
+When you include the '-s ha' option, you will be promoted right away for the load balancer address. As shown in the figure below, we are providing the 'internal' IP address. This address is used by the Kubernetes services in the nodes to talk to each other. Using a public IP addresses addds extra layers that are not needed. 
+
+This process will take several minutes as it will install Kubernetes as well as the Application Administration toolset. Once it is finished, you should see something similar as shown below:
```
dexhorthy, line 240: is there a reason why `Application Administration` is title case?

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+![Output](/images/guides/kots/ha-cluster-install-output.png)
+
+Highlighted in red in the screenshot above are first the Kots Admin URL and password to login. Also highlighted are the commands to run on other nodes to join the cluster. In this exercise we are not going to run these commands yet and instead continue deploying the application. The commands to join the cluster are also available in the Kots Admin UI, which we will cover later in this guide.
```
dexhorthy, line 244: What do you think about removing `first` in the opening sentence here?

```suggestion
Highlighted in red in the screenshot above are the Kots Admin URL and password to login. Also highlighted are the commands to run on other nodes to join the cluster. In this exercise we are not going to run these commands yet and instead continue deploying the application. The commands to join the cluster are also available in the Kots Admin UI, which we will cover later in this guide.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+Next, we are going to log in to Kots Admin to complete the install and deploy the application. Since this guide assumes you have a level of familiarity with Kots, we are not going to go into detail on every step to get the application deployed.
```
dexhorthy, line 246: What do you think about linking to a quickstart guide for those who are not familiar yet?

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+![Log-In](/images/guides/kots/ha-cluster-log-in.png)
+
+Log in to the Aplication Administration Console using the address and password from the install. Once you have logged in using the provided password, you will need to provide a license. The license is generated in the Vendor Portal and this guide assumes you already know how to do this. There are no options that need to be enabled or turned on in order to support an HA cluster, so if you already have a devlopment license, it should work for this exercise.
```
dexhorthy, line 250: I think since we just mentioned above this section that "you should already know how to do this", we might be able to omit it here. What do you think?

Also - small typo here 
```suggestion
Log in to the Aplication Administration Console using the address and password from the install. Once you have logged in using the provided password, you will need to provide a license. The license is generated in the Vendor Portal and this guide assumes you already know how to do this. There are no options that need to be enabled or turned on in order to support an HA cluster, so if you already have a development license, it should work for this exercise.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+### Verify the Deployment
+
+To see everything that has been deployed, ssh into one of the nodes and run:
+
+```shell
+kubectl get pods --all-namespaces -o wide
+```
+
+
+The output will show you all the pods running on all namespaces and on which node. Since we only have a one node cluster, all pods should be running on node 1. Check the status of the pods to make sure there are no issues.
```
dexhorthy, line 275: This first sentence sounds a little awkward to me -- particularly the last bit about `on which node`

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+![AllPods](/images/guides/kots/ha-cluster-all-pods-one-node.png)
+
+If you are following along, you should be able to access the application by browsing to http://<ip-address-of-ha-proxy>. The sample app uses flask 'routes' to call various methods to interact with Postgres. To test if it can write to the database, use the  `/sql-check`   route which will check connection to the default database and list the databases are avaialble to the 'postgres' user. To write to the database, first use the `/sql-create` route which creates a database (appdiretdb) with a table (tblrecords) in it. Finally, use the `/sql-add` route to add a row the table with the timestamp. Once the database and table are created, simply use this route to add more records to the database.
```
dexhorthy, line 279:
```suggestion
If you are following along, you should be able to access the application by browsing to http://<ip-address-of-ha-proxy>. The sample app uses flask 'routes' to call various methods to interact with Postgres. To test if it can write to the database, use the  `/sql-check`   route which will check connection to the default database and list the databases that are avaialble to the 'postgres' user. To write to the database, first use the `/sql-create` route which creates a database (appdiretdb) with a table (tblrecords) in it. Finally, use the `/sql-add` route to add a row the table with the timestamp. Once the database and table are created, simply use this route to add more records to the database.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+To verify that the data is in fact being written to Postgres, install pgAdmin and create a connection to the database using the Load Balancer's IP address. Once conneted, you should see the database and table. To retrieve records you can run a new query like this:
```
dexhorthy, line 281: What do you think about making this more general? E.g. instead of requiring `pgAdmin`, why not suggest "the postgres client of your choice"? It might help as well to include a basic `psql` command for folks to connect from a command line, and show them how to get the connection details for the DB embedded in appdirect?

I am not sure this is necessary, just thinking out loud. I guess it depends how many folks end up following along w/ AppDirect vs using their own app.

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+![pgAdmin](/images/guides/kots/ha-cluster-pg-admin.png)
+
+### Adding Remaining Nodes to Cluster
+
+Under Cluster Management you can view the status of the embedded cluster. At this time, it only has one node. 
+
+
+![ClusterManagement](/images/guides/kots/ha-cluster-cluster-mgmt.png)
+
+To get the commands to run on the other nodes, click on the 'Add Node' button. From here you can retrieve the command to use on Worker and Master nodes. For our exercise we will add Master nodes to give us a proper HA solution.
```
dexhorthy, line 296: I think "HA Cluster" makes more sense than "HA solution" here, since the cluster is only a piece of the solution, but this one is super subjective so it's your call.

```suggestion
To get the commands to run on the other nodes, click on the 'Add Node' button. From here you can retrieve the command to use on Worker and Master nodes. For our exercise we will add Master nodes to give us a proper HA cluster.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+![ClusterManagement](/images/guides/kots/ha-cluster-add-node-cmd.png)
+
+Copy the command and then ssh into nodes 02 & 03 and paste the command and hit enter. This should run a process similar to the initial install of the cluster and will take a few minutes. Once it is finished you should see something similar to the output below:
```
dexhorthy, line 300: There's a lot of and X and Y and Z steps in this first sentence, is there a way to make it flow better?

content/vendor/guides/ha-guide.md
cremerfc: I like that

content/vendor/guides/ha-guide.md
cremerfc: How about something like.... Once you make changes to the Kubernetes Installer, make sure to promote the installer to the same channel(s) as your application installers. In this exercise, the Kubernetes installer was promoted to the 'Unstable' channel, since that is where the application is also promoted to.

content/vendor/guides/ha-guide.md
cremerfc: I guess I was thinking first of "will install the Application Admin console", but then added the tooling part, which I guess makes it 'a set of things' versus 'one specific item I am calling by its proper name"... so should be lower case

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+```shell
+kubectl get pods --all-namespaces -o wide
+```
+
+The output should show that other than the default namespace, there are pods running on all three nodes.
+
+# Part II: Testing Scenarions & Troubleshooting
+
+There are various testing scenarios for an HA Cluster. For starters we are focusing on data retention/loss scenario. The sample application is already configured to write to the embedded Postgres StatefulSet.
```
dexhorthy, line 321: I think this needs a tweak but I don't know what. Tried with `the` lemme know what you think 


```suggestion
There are various testing scenarios for an HA Cluster. For starters we are focusing on the data retention/loss scenario. The sample application is already configured to write to the embedded Postgres StatefulSet.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+## Pre-Requisites
+
+This section assumes you followed the above steps in order and are using the sample application. Since we are testing data retention, make sure to use the routes described in the [Verify the Deployment](#verify-the-deployment) section.
+
+To verify that data is not lost, use pgAdmin as described above. Simply run the query again, and should be presented with the same results. You should also have a separate terminal in which to run:
```
dexhorthy, line 327: Again I think it would be great to provide plain `psql` commands in addition to suggesting the use of pgAdmin. I don't recall the table names or anything, but maybe something like this?

```ssh
psql --host $HAPROXY_ADDRESS --user ... --password 'SELECT * FROM some_table'
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+```shell
+watch -d 'kubectl get pods --all-namespaces -o wide'
+```
+
+The above command will display all of the pods in the cluster and on which node they are running on. By using `watch -d` the output will automatically get refreshed so you can watch pod activity as you delete pods and node.
```
dexhorthy, line 333: First sentence reads strange, might need another look.

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+## Deleting a Pod
+
+If you followed along with the steps and sample application, you should see that the Postres pod is running on Node 01. In this test, we are going to delete the Postgres pod. The expected result is that a new pod will be scheduled (possibly on another node) and all the data in the database will be retained and not lost.
+
+To delete the Postgres pod deployed with the sample app, run the following command:
+
+```shell
+kubectl delete pod app-direct-postgresql-0
+```
+This command will schedule a new pod and then terminate the existing one. Verify that the new pod is up and running and check that the data is still there by using pgAdmin. You can also use the various routes in the sample app to verify database connectivity from the app.
```
dexhorthy, line 344: ditto re: psql

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+- As described in the [Kubernetes Documentation](https://kubernetes.io/docs/concepts/architecture/nodes/#condition), there is a default 5 minute timeout duration. This means that there will be 5 minutes between the time the node goes down and that Kubernetes will finally decide that the node is not coming back and will schedule pods on this node to another node.
+
+To simulate a node becoming unresponsive or simply crashed, you can simply stop the VM corresponding to the node. You are likely not to see much change in the pods during the first five minutes, other than some pods erroring out. After 5 minutes or so, you will start to see pods being terminated and scheduled on the remaining nodes. 
+
+In this scenario, the downtime for Postgres was about 6 - 7 minutes. This includes the Kubernetes time out of 5 minutes, 1 minute or less for ekco to detect the node being evicted, and the remainig time is for Postgres to start up on a new node.
```
dexhorthy, line 356:
```suggestion
In this scenario, the downtime for Postgres was about 6 - 7 minutes. This includes the Kubernetes time out of 5 minutes, 1 minute for EKCO to delete the node and drain the data, and some remaining time is for Postgres to start up on a new node.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+
+## Troubleshooting
+
+While the above process has been tested several times and tested succesfully the retention of data, there were a couple of issues that you may run into and are documented below
```
dexhorthy, line 361: this part reads a little strangely to me, `tested several times and tested successfully the retention of data`

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+```
+This appeared to be more of an OS issue, so I deleted this pod. When the new one was scheduled all was fine.
+
+### Postgres unable to mount to data directory
+
+In some occassions when you shut down the node that Postgres is running on, you may run into a situation where the new pod that gets scheduled on one of the remaining nodes is not able to come up due to volume mount issues. Running `kubectl describe` on the Postgres pod may show an error like the following:
+
+![PostgresDesribe](/images/guides/kots/ha-cluster-postgres-describe.png)
+
+The first thing to check is to ensure you have followed the [configuring ekco] section. Ekco helps with ensuring that any resources that were running on the stopped node are properly deleted and moved to a different node. Without ekco, the claim on the PVC associated to the Postgres pod that was running on the stopped node sometimes doesn't get properly cleared and when the new pod gets scheduled on another node, it is not able to lay a claim to the volume.
```
dexhorthy, line 379: missing link here I think? `[configuring ekco]`

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+To simulate a node becoming unresponsive or simply crashed, you can simply stop the VM corresponding to the node. You are likely not to see much change in the pods during the first five minutes, other than some pods erroring out. After 5 minutes or so, you will start to see pods being terminated and scheduled on the remaining nodes. 
+
+In this scenario, the downtime for Postgres was about 6 - 7 minutes. This includes the Kubernetes time out of 5 minutes, 1 minute or less for ekco to detect the node being evicted, and the remainig time is for Postgres to start up on a new node.
+
+
+## Troubleshooting
+
+While the above process has been tested several times and tested succesfully the retention of data, there were a couple of issues that you may run into and are documented below
+
+### Postgres pod will not come up after deleting the pod
```
dexhorthy, line 363: I love the inclusion of a `Troubleshooting` section, more of our guides should have this! I noticed there's a little inconsistency in the cases, this first one uses `I` and `would` to describe a thing that happened, where as the second one is in the second/person with `you` and `may` to describe potential cases. What do you think about consolidating to one style?

content/vendor/guides/ha-guide.md
dexhorthy: This statement about how the Claim is cleared sounds right but I don't know enough about EKCO to say for sure. Let's check w/ core on the phrasing here?

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+![PostgresDesribe](/images/guides/kots/ha-cluster-postgres-describe.png)
+
+The first thing to check is to ensure you have followed the [configuring ekco] section. Ekco helps with ensuring that any resources that were running on the stopped node are properly deleted and moved to a different node. Without ekco, the claim on the PVC associated to the Postgres pod that was running on the stopped node sometimes doesn't get properly cleared and when the new pod gets scheduled on another node, it is not able to lay a claim to the volume.
+
+### Other Persistent Data Issues
+
+The default configuration of the Kurl installer includes the [rook add-on](https://kurl.sh/docs/add-ons/rook), which creates and manages a [Ceph cluster](https://docs.ceph.com/docs/master/rados/) along with a storage class for provisioning Persistent Volume Claims (PVCs). If you are running into other issues with persisitent storage, following are some helpul commands to check the health of ceph in your cluster. Keep in mind that the screenshots below show a `healthy` cluster that has all nodes up and running.
```
dexhorthy, line 383:
```suggestion
The default configuration of the Kurl installer includes the [rook add-on](https://kurl.sh/docs/add-ons/rook), which creates and manages a [Ceph cluster](https://docs.ceph.com/docs/master/rados/) along with a storage class for provisioning Persistent Volume Claims (PVCs). If you are running into other issues with persistent storage, following are some helpful commands to check the health of ceph in your cluster. Keep in mind that the screenshots below show a `healthy` cluster that has all nodes up and running. 
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+### Other Persistent Data Issues
+
+The default configuration of the Kurl installer includes the [rook add-on](https://kurl.sh/docs/add-ons/rook), which creates and manages a [Ceph cluster](https://docs.ceph.com/docs/master/rados/) along with a storage class for provisioning Persistent Volume Claims (PVCs). If you are running into other issues with persisitent storage, following are some helpul commands to check the health of ceph in your cluster. Keep in mind that the screenshots below show a `healthy` cluster that has all nodes up and running. 
+
+All related pods are runnning in the rook-ceph namespace, and a good place to start to troubleshoot any issues. By running `kubectl get pods -n rook-ceph` you should get an output similar to the one below:
```
dexhorthy, line 385:
```suggestion
All related pods are runnning in the rook-ceph namespace, and a are good place to start to troubleshoot any issues. By running `kubectl get pods -n rook-ceph` you should get an output similar to the one below:
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+![PostgresDesribe](/images/guides/kots/ha-cluster-rook-ceph-pods.png)
+
+One thing to note here is that you will see pods running on all three nodes. Orchestrating all of the data accross all nodes is the `rook-ceph-operator-...` pod. To troubleshoot issues, we'll need to exec into this pod in order to run some commands. For example, to exec into the pod on the screenshot above, I would run `kubectl exec -it rook-ceph-operator-6fbfdccc57-p477z`. As you exec into the pod, you may see a few errors but you can igonre them.
```
dexhorthy, line 389: This seems dangerous `As you exec into the pod, you may see a few errors but you can igonre them.` -- do you have an example of what kind of errors can be ignored? (also typo in ignore)

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+
+To get another look at the OSDs, run `ceph osd tree`
+
+![OSDTree](/images/guides/kots/ha-cluster-osd-status.png)
+
+## Next Steps
+
+The guide covers a basic 3 node HA cluster. However, there are other layers of complexity that could be added depending on your use case:
+
+- Add worker nodes to the cluster and run same tests.
```
dexhorthy, line 409:
```suggestion
- Add worker nodes to the cluster and run the same tests.
```

content/vendor/guides/ha-guide.md
```diff
@@ -0,0 +1,411 @@
...
+- Set up https access 
+- Setting up Postgres in HA mode and compare down time against a single Postgres deployment.
```
dexhorthy, line 411: Should this be `Set` to match the phrasing of the first two bullets?

content/vendor/guides/ha-guide.md
cremerfc: No I am thinking there is really no point to even mention this... so I have removed that bullet point.

content/vendor/guides/ha-guide.md
cremerfc: thanks!

content/vendor/guides/ha-guide.md
cremerfc: OK I kinda rewrote this whole section.

content/vendor/guides/ha-guide.md
cremerfc: I've replaced this with "The main purpose of the load balancer is to give us a single point of access." I think that is really what the load balancer gives us.

content/vendor/guides/ha-guide.md
cremerfc: I went through the guide and ensured it consistently refers to it as 'HAProxy'

content/vendor/guides/ha-guide.md
cremerfc: I changed this to:

This process will take several minutes as it will install a Kubernetes 'stack' that will include the add-ons defined in the Kubernetes Installer, including Kots. Once it is finished, you should see something similar as shown below:

content/vendor/guides/ha-guide.md
cremerfc: I broke it down into two sentences:

The output will show you all the pods running on all namespaces. The output also shows you the node that each pod is currently running on, which will come in handy during this exercise.

content/vendor/guides/ha-guide.md
cremerfc: I met you half way... I added a bit about using 'Postgres client of your choice' but also noted that the screenshots are from pgAdmin.

content/vendor/guides/ha-guide.md
cremerfc: I modified this section and I think it flows better.

content/vendor/guides/ha-guide.md
cremerfc: I feel that if I include the command I should include sample output, which I don't and rather get this guide out. Maybe a 2.0 version is to include examples with psql.

I did modify this section to be more general.

content/vendor/guides/ha-guide.md
cremerfc: removed that sentence, because I already explained what the Kubectl command does and shows. Only the explanation of the watch command is needed IMO

content/vendor/guides/ha-guide.md
cremerfc: Again, want to get this out the door, changed this whole section to be more generic. Happy to revisit this and add psql in the future.

content/vendor/guides/ha-guide.md
cremerfc: I rephrased it to "As you go through this exercise, you may run into some of the issues below:"

content/vendor/guides/ha-guide.md
cremerfc: I updated the entire section, hopefully it is more consistent

content/vendor/guides/ha-guide.md
cremerfc: I just removed the sentence :)

content/vendor/guides/ha-guide.md
cremerfc: yes, done

content/vendor/guides/ha-guide.md
cremerfc: Good catch on the link. I ended up changing the whole Troubleshooting section and ended up removing that part.

content/vendor/guides/_index.md
```diff
@@ -43,6 +43,11 @@ The KOTS CI/CD Integration guide is an advanced workflow that allows you to use
 This guide is intended as a lightweight checklist to ensure you have explored all the features of KOTS and integrated the ones that will help ensure your end users are most likely to be successful deploying and running your KOTS application
 
 
+###### [Deploy an HA Cluster](/vendor/guides/ha-guide)
+
+This guide is intented to guide the reader in setting up a basic 3 master node Kubernetes cluster in an [Embedded Cluster](/vendor/guides/quickstart) scenario for learning and testing purposes.
```
marccampbell, line 48: Could you replace "master" with a preferred term such as "primary"?  See the contributing guide for additional recommendations

PR #272, opened 2020-09-09
Add note to LicenseFieldValue about allowed vals

robbiet480:
Had to dig through source of the template function as well as the REST API to figure out allowed values, figured I'd save others work by documenting them.


robbiet480:
Added a note in the `LicenseFieldValue` source requesting a docs update whenever the function changes: https://github.com/replicatedhq/kots/pull/1079

PR #284, opened 2020-09-19
Reword kotsadm 1.19.3 release note

areed:


markpundsack (approved):
Was the prefer thingy material to mention? Approved as-is though.

areed:
I think it's too low-level for release notes

PR #289, opened 2020-09-24
Update Airgap Existing Guide

aj-jester:
* Updated the kURL addon URL to a new one which includes EKCO. I don't want to terminate this setup because it was time consuming to set it up, rather I just want to stop the nodes when I'm not using them.

* Updated the `plain-registry.yaml` URL to the replicated-automation repo so it reads a little easier than the hash. PR here https://github.com/replicatedhq/replicated-automation/pull/15

* Showed an example of how success vs. error looks like when testing the registry. Also the screenshot of the docker config was small so it wasn't clear how the syntax was supposed to be, so instead added the JSON blob.

### DIFF
--- a/content/vendor/guides/airgapped-existing-cluster.md
+++ b/content/vendor/guides/airgapped-existing-cluster.md
@@ -141,17 +141,17 @@ echo ${CLUSTER_PRIVATE_IP}
 ```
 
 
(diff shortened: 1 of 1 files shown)


dexhorthy (approved):
@aj-jester this looks great, let's get it in 🙂

PR #319, opened 2020-11-03
kind: ConfigValues example

divolgin:


content/kotsadm/installing/automating.md
```diff
@@ -33,7 +34,16 @@ This file will be:
...
+metadata:
+  name: myapp
+spec:
+  values:
+    app_name:
+      value: MyApp
+    gitlab_app_secret:
+      valuePlaintext: password123
+    primary_domain:
+      default: somebigbank.org
```
markpundsack, line 46: That seems like a weird default. Should this be `value` instead?

markpundsack: While you could, of course, include defaults in here, presumably the this config.yaml would be merged with the upstream so defaults would be filled in.

divolgin: This just shows that  `default` can be set here.  There's an example with `value` already.

AmberAlston:
@markpundsack Looking at this PR since it's getting a little old. Do you believe the PR needs to be modified, or thought about more, before we move it forward?

markpundsack:
@AmberAlston It's fine as-is. My worry is that someone sees that an interprets that if they want to override a default value, they should just change the value to whatever they want, and I think in reality, they should change it from a `default` to a `value`. You can leave defaults there, but you shouldn't change defaults there. Maybe that just needs clarification? At any rate, this could be merged and then iterated upon.

AmberAlston:
@divolgin I committed a very minor change to fix "An" to "an" for nav consistency. Per the comments above, you can merge this if you want to push it forward.

PR #337, opened 2020-11-19
release notes for kots 1.23.1

divolgin:


salahalsaleh (approved):

markpundsack (approved):

markpundsack:
Approved, but I wonder if we can be a little more informative of what the problem is. Like is there a way we can describe what the impact was? As I understand it, the bug is that we stopped respecting requests to exclude things from backup...

PR #340, opened 2020-11-20
[1.24.0] Docs for kots remove command

divolgin:


content/kots-cli/remove.md
```diff
@@ -0,0 +1,32 @@
...
+linktitle: "remove"
+title: kots remove
+weight: 90120
+---
+
+Remove application reference from Admin Console.
+Multiple applications can be installed using the `Add a new application` button in Admin Console or using the `kots install` command.
+All installed applications will appear in Admin Console.
+The `kots remove` command can be used to remove them.
+Note that deployed application will not be removed from the cluster, only it's reference will be removed from Admin Console.
```
markpundsack, line 12:
```suggestion
Note that the deployed application will not be removed from the cluster; only it's reference will be removed from Admin Console.
```

content/kots-cli/remove.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Usage
+```bash
+kubectl kots remove [app-slug]
+```
+* _Replace `[app-slug]` with the slug for the KOTS application that should be removed (required)._
```
markpundsack, line 18: You usually don't need to say "replace". People know how CLI docs work, right? Just say "app-slug is..."

content/kots-cli/remove.md
```diff
@@ -0,0 +1,32 @@
...
+* _If the KOTS application has been packaged by Replicated Vendor, the `--license-file` flag must be provided._
```
markpundsack, line 19: Why is the license file required on remove? Is that just for a safety check? Seems burdensome. Also, how would a Kots application not be packaged by Replicated Vendor? Why is Vendor capitalized? Needs a "the".

content/kots-cli/remove.md
```diff
@@ -0,0 +1,32 @@
...
+* _Provide `[flags]` according to the table below_
+
+This command supports all [global flags](/kots-cli/global-flags/) and also:
+
+
+| Flag                 | Type | Description |
+|:----------------------|------|-------------|
+| `--force` |  bool  |  if application was deployed, the command will fail unless the force flag is specified |
```
markpundsack, line 27:
```suggestion
| `--force` |  bool  |  Remove the reference even if the application has already been deployed. |
```

content/kots-cli/remove.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Example
+```bash
+kubectl kots remove sentry -n default
```
markpundsack, line 31: The text above says a license is required for nearly all instances, so maybe it should be provided here? (or drop that requirement)

divolgin: that was just copy/paste stuff from another doc

PR #350, opened 2020-12-10
a few small changes

dexhorthy:
The default YAML no longer includes the replicas: 1 line in it, because
that was triggering a linter warning. I've updated the guide to handle that better


AmberAlston:
@dexhorthy Are you still wanting to move forward with this suggested change? If so, @aj-jester are you able to review as requested?

AmberAlston:
@dexhorthy Are you still wanting to move forward with this update?

AmberAlston:
Kicking this PR again in case you want to move forward with it. If not, we're going to close it as stale in the near future.

PR #364, opened 2020-12-29
[1.27.0] identity service admin console RBAC section

salahalsaleh:
## Updated Screenshot
Image 2020-12-29 at 2 24 41 PM

## New Screenshots
Image 2020-12-29 at 10 39 35 AM

Image 2020-12-29 at 2 15 53 PM


AmberAlston (commented):
I made one small copy edit for readability .

Question - How is a user meant to understand what these  “pre-defined Admin Console roles” are? Is there anywhere in the docs now that we describe these roles & their respective restrictions? 

Nice-to-have - We made the screenshot above this (under configuration) outdated. It would be nice to update that to show what a v1.27 user would see now (with RBAC Add a Group choice ahead of Advanced Options  & Save)

PR #365, opened 2020-12-29
[1.27.0] Identity Service Application RBAC - vendor

salahalsaleh:
Image 2020-12-29 at 11 44 22 AM

Image 2020-12-29 at 11 44 27 AM


AmberAlston (approved):
This looks good as-is. I expect we may find ways to continue to improve it once customers start leveraging, and we can see where any confusion points in the docs may be.

PR #375, opened 2021-01-08
[1.28.0] docs for 'kots get app' CLI

divolgin:


AmberAlston (changes requested):
The new page content looks good but `get` is not in an ideal spot on the nav. It shouldn't break up restore commands 
image

PR #378, opened 2021-01-12
Admin console snapshots

jgruica:


AmberAlston (commented):
For the content now under Admin Console, the sub-nav is in alphabetical order which doesn't feel right. It makes more sense to be in the order of expected operations, so I suggest changing ordering to: Configuring Backup, Snapshot Destinations, Schedules, Snapshot Restore (changed from Configuring Restore), Disaster Recovery, Troubleshooting

Nice-to-have: under vendor snapshots overview page, let's link "Admin Console" wording back to the Admin Console snapshot section.

AmberAlston (approved):

PR #382, opened 2021-01-19
admin console - types of snapshots

salahalsaleh:
Image 2021-01-19 at 3 18 31 PM

Image 2021-01-19 at 3 18 35 PM


AmberAlston (changes requested):
Minor feedback 
- Would prefer “Snapshot Types” to “Types of Snapshots”. It’s shorter and matches existing Destinations convention.
- Let’s remove the sentence “If you only need a partial backup of just application volumes and manifests for rollbacks, use Application Snapshots” from Instance snapshots section. Given we’d like to force a behavior change on the path to deprecating it, this sentence sounds like we actively encourage it for some use cases. I think the product UI this content came from will also change soon.

AmberAlston (approved):

PR #415, opened 2021-02-22
Removed words that imply expired license can stop running applications

tdensmore:
This was probably copy pasta.


AmberAlston (approved):
Thanks @tdensmore !

PR #440, opened 2021-03-19
1.35.0 Digest support on Kots apps

jala-dx:


AmberAlston (changes requested):
The place this was inserted in the page broke up a paragraph where the second part was a direct reference to the first part. That section now reads as if "this is useful and recommended" is in reference to the text you added, which is confusing. The text is fine but it's not in the right section on this page as-is. A note call out at the top or bottom of that section may make more sense.

AmberAlston:
image

AmberAlston (approved):
Looks good with changes

PR #441, opened 2021-03-24
Fix the registry ip that sounds like a Domain name

jala-dx:


salahalsaleh (commented):
both a hostname or an ip are valid, and i think "host" is a  more generic term in this case.

content/vendor/packaging/workspace.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
```
salahalsaleh, line 1: this file was probably pushed unintentionally (cleanup)

AmberAlston:
@jala-dx Can we close this given its age?

PR #485, opened 2021-05-26
update KOTS install command

marydavis:
Doc location: https://kots.io/kotsadm/installing/airgap-packages/#kots-install

Removed `--kotsadm-namespace app-name` as it is not in the latest [install video](https://www.youtube.com/watch?v=OLjwKSuRCXM) ~and it prompts for the namespace to deploy to once the command is run~.

Added `/app-name` to the end of ` --kotsadm-registry private.registry.host` to match previous push to registry command.

### DIFF
--- a/content/kotsadm/installing/airgap-packages.md
+++ b/content/kotsadm/installing/airgap-packages.md
@@ -37,8 +37,7 @@ These credentials will be used to pull the images, and will be automatically cre
 
 ```shell
 kubectl kots install app-name \
-  --kotsadm-namespace app-name \
-  --kotsadm-registry private.registry.host \
+  --kotsadm-registry private.registry.host/app-name \
   --registry-username ro-username \
   --registry-password ro-password
 ```


salahalsaleh:
@marydavis the `--kotsadm-namespace` flag is to indicate the registry namespace to use for kotsadm images in airgapped mode, not the namespace in which kotsadm will be deployed. for that, you can use the `--namespace` flag, if provided, you won't be prompted for the namespace.

salahalsaleh:
also @marydavis, the flag does exist in the video (7:27), here's a screenshot 
Image 2021-06-01 at 9 41 53 AM

salahalsaleh:
also, the official docs mention that the `--kotsadm-namespace` flag should be used in airgapped mode for the kots install command in existing clusters: https://kots.io/kotsadm/installing/airgap-packages/#kots-install

marydavis:
@salahalsaleh oh ok, so the instruction video just uses a shortcut and puts the namespace as part of the `--kotsadm-registry` flag but that is not the recommended way?

salahalsaleh:
@marydavis both formats are valid, but updating the docs will take more than just updating what's in this PR. plus, some customers reference our docs in their docs, and this will cause them to have inconsistencies and require them to update their docs. since the current format is valid, and has been around for a long time, i would suggest to leave it as is.

marydavis:
@salahalsaleh that is actually how this PR came about in the first place as we link to this install doc too :) We had received this feedback from a customer and just thought it would be helpful and clear up some confusion. I understand the reason to not change it so we will just add a note to our docs and I will close the PR. Thanks for reviewing and the explanation.

salahalsaleh:
@marydavis sounds good, thanks!

PR #488, opened 2021-05-28
Update installing-embedded-cluster.md

JustinN-Replicated:
The old URL https://kurl.sh/docs did not resolve


JustinN-Replicated:
Worth mentioning that https://kurl.sh/docs loaded a blank white page, but https://kurl.sh/docs/ does correctly loads/redirects to https://kurl.sh/docs/introduction; I had changed the link  to https://kurl.sh/docs/introduction (which resolves) but it could also be changed to https://kurl.sh/docs/

tdensmore:
@JustinN-Replicated just a comment: sometimes these requests get lost unless you explicitly add reviewers to the PR.

JustinN-Replicated:
@jgruica could I have you review the small doc change since we discussed? Should take just a second :)

AmberAlston (approved):

PR #513, opened 2021-06-25
code copy + blockquote

breezybaldwin:
customer suggested wording improvement


AmberAlston:
@breezybaldwin is this still in-progress or ready for review? If ready for review, can you add a reviewer please?

AmberAlston:
@breezybaldwin Per our chat, let me know when you think this is actually reviewable again, i.e. those elements that looked weird are fixed. 

@dexhorthy it's unclear what pattern we were following for adding a code copy button vs not on the code blocks. I don't have the history on this change from last fall as it pre-dates me.

dexhorthy:
@AmberAlston I think there are many pre-formatted blocks that show things that are not suitable for copying, for example expected output from commands. I haven't previewed this but the idea would be that documentation writers could either opt-in or opt-out of having a copy button on a `pre` or `code` block.


For example, [the code blocks here](https://kots.io/vendor/guides/persistent-datastores/#adding-the-secret-and-statefulset) could have a copy button, a block like this one should not, at least not the output part. 

Screen Shot 2021-06-27 at 12 44 05 PM


We could even discuss the option of allowing the documentation writer to specify separately what gets copied (just the `kubectl get ...` part in this case).

@breezybaldwin is the "copy code" button available to play with on your PR or do we need to add something to an existing documentation page to see it in action?

AmberAlston:
@dexhorthy The approach you outlined make sense. My question stemmed from an initial review of this PR, and wanting to understand what the roll-out plan was to ensure it's applied throughout the docs. As-is, it was applied on a couple of pages. Looking at those pages, it was applied to code blocks that made sense to copy and some code-blocks that it didn't (or at least not everything in the current code block). If we roll out a change like this to the docs, I want to ensure we're being consistent so users aren't left wondering why they can copy on some pages, but not others where we also show useful-to-copy code blocks. It feels like the code block change really needs to paired with an audit to go through sections of the website and apply the right formatting where appropriate. From my convo with @breezybaldwin it was unclear who's taking responsibility for driving the change throughout the KOTS docs. 

Your can see the page I was looking at here: https://deploy-preview-513--kotsio.netlify.app/vendor/guides/existing-cluster/

breezybaldwin:
@AmberAlston I'm happy to own this in a few months when I have a larger team and we aren't as busy with a full rebrand

AmberAlston (changes requested):
Needs a content audit to ensure code block is appropriately applied to various docs sections (could be tackled per section)

AmberAlston:
FYI on this open PR @replicatedhq/replicated-docs Something to think about in the new docs site execution.

samantha-replicated:
Adding myself as a reviewer so I'll get notified from GH and able to follow along :) Not because I need to sanction it for merging.

jonquil2002:
@breezybaldwin @dexhorthy I'm closing this PR as the work was only partially completed last year and the work that remains is extensive. If it still makes sense to do this work on the new docs site, then let's open a Shortcut story so that we can address it there. Thanks!

PR #528, opened 2021-07-20
kURL aligning minimum requirements

jdewinne:


AmberAlston:
This was approved back in July but never merged. @jdewinne or @JustinN-Replicated please merge if it's still a change we want to make.

PR #532, opened 2021-07-29
document generate-manifests command

morningvera:
[CH36021](https://app.clubhouse.io/replicated/story/36021/a-cli-command-or-a-flag-for-the-kots-pull-command-to-generate-only-the-admin-console-components-as-kustomizable-yaml-manifests)

### DIFF
--- a/content/kots-cli/admin-console/generate-manifests.md
+++ b/content/kots-cli/admin-console/generate-manifests.md
@@ -5,13 +5,14 @@ title: kots admin-console generate-manifests
 weight: 90130
 ---
 
-If you’d rather use kubectl or another workflow to deploy to a cluster, run `kots admin-console generate-manifests` to create a directory on the workstation containing the KOTS Admin Console manifests. This workflow enables additional customization of the Admin Console before deploying.
+Running this command will create a directory on the workstation containing the KOTS Admin Console manifests. These assets can be used to deploy KOTS to a cluster through other workflows, such as kubectl, to provide additional customization of the Admin Console before deploying.
 
 ### Usage
 ```bash
 kubectl kots admin-console generate-manifests [flags]
 ```
-* _Provide `[flags]` according to the table below_
+
+This command supports the following flags:
 
 | Flag | Type | Description |
 |:-----|------|-------------|
--- a/content/kots-cli/admin-console/generate-manifests.md
+++ b/content/kots-cli/admin-console/generate-manifests.md
@@ -2,7 +2,7 @@
 date: 2021-07-29
 linktitle: "generate-manifests"
 title: kots admin-console generate-manifests
-weight: 90110
+weight: 90130
 ---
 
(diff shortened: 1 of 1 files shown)


content/kots-cli/admin-console/generate-manifests.md
```diff
@@ -0,0 +1,29 @@
+---
+date: 2021-07-29
+linktitle: "generate-manifests"
+title: kots admin-console generate-manifests
+weight: 90110
```
sgalsaleh, line 5: there's already a command with this weight. i would suggest usin `90130`

morningvera: done, good catch!

content/kots-cli/admin-console/generate-manifests.md
```diff
@@ -0,0 +1,29 @@
...
+weight: 90130
+---
+
+If you’d rather use kubectl or another workflow to deploy to a cluster, run `kots admin-console generate-manifests` to create a directory on the workstation containing the KOTS Admin Console manifests. This workflow enables additional customization of the Admin Console before deploying.
+
+### Usage
+```bash
+kubectl kots admin-console generate-manifests [flags]
+```
+* _Provide `[flags]` according to the table below_
```
JustinN-Replicated, line 14: I think we should match the format of other CLI commands; also are global flags supported? 
Replacement could be:
This command supports all [global flags](/kots-cli/global-flags/) and also:

morningvera: That line was actually copied from the `kots pull` docs but I'm fine with removing it. I didn't include the part about global flags because they're not supported currently. But definitely open to matching the format of other docs!

I'm not sure `kots pull` really supports global flags either (even though it says so). Looking at the code, that CLI command doesn't seem to read any of the global flags.

JustinN-Replicated: Totally makes sense; I wasn't sure about the globals part at all but figured to include just in case; could likely be adjusted to:
This command supports the following flags:

content/kots-cli/admin-console/generate-manifests.md
```diff
@@ -0,0 +1,29 @@
+---
+date: 2021-07-29
+linktitle: "generate-manifests"
+title: kots admin-console generate-manifests
+weight: 90130
+---
+
+If you’d rather use kubectl or another workflow to deploy to a cluster, run `kots admin-console generate-manifests` to create a directory on the workstation containing the KOTS Admin Console manifests. This workflow enables additional customization of the Admin Console before deploying.
```
JustinN-Replicated, line 8: This could maybe be reworked to align a bit with our contribution guidelines: 
Running this command will create a directory on the workstation containing the KOTS Admin Console manifests. These assets can be used to deploy KOTS to a cluster through other workflows, such as kubectl, to provide additional customization of the Admin Console before deploying.

morningvera: Ah, I was trying to figure out a good way to reword that first sentence to remove the "you" but didn't come up with anything. Thanks for the suggestion!

JustinN-Replicated (approved):
Looks great!

PR #548, opened 2021-08-10
Jeffgolden/ch36524/docs for user api tokens

jeffreygolden:


content/vendor/cli/tokens.md
```diff
@@ -0,0 +1,31 @@
...
+date: 2021-08-09
+linktitle: "Tokens"
+title: Using Vendor API Tokens
+weight: 90105
+draft: false
+---
+
+Using the Vendor CLI requires tokens for authentication.  Replicated supports two types of tokens: Team tokens and User tokens.
+
+Tokens may only access a portion of the Vendor API. Administrative tasks such as adding a new user, creating a new token, or changing a password cannot be performed with token authentication.  These administrative tasks must be performed by a logged-in user.
```
JustinN-Replicated, line 11: Can we clarify which portions of the API  in a succinct way? I am not sure if this is accurate, but for example we could link to the [resource names](https://help.replicated.com/docs/vendor-rbac/resource-names/) or something. If that is the case, I think it would be good to be a bit more specific.

PR #553, opened 2021-08-16
Added the 'placeholder' annotation to the reference guide.

tdensmore:
I have added the "placeholder" annotation to the reference documentation. 
Hopefully, I got the syntax correct, but I will need extra eyeballs on that (since I could not find any documentation on it).
This was requested by Wallaroo.

### DIFF
--- a/content/reference/v1beta1/backup.md
+++ b/content/reference/v1beta1/backup.md
@@ -10,7 +10,7 @@ weight: 7
 A Backup resource in an application causes the Admin Console to enable [snapshots](/vendor/snapshots/overview/) for the application. 
 This resource is [fully documented](https://velero.io/docs/v1.5/api-types/backup/) on velero.io.
 
-This resource supports the [KOTS optional resources](/vendor/packaging/optional-resources/) annotations.
+This resource supports the [KOTS optional resources](/vendor/packaging/include-resources/) annotations.
 
 ```yaml
 apiVersion: velero.io/v1
--- a/content/release-notes/1.9.0.md
+++ b/content/release-notes/1.9.0.md
@@ -7,7 +7,7 @@ title: "1.9.0"
 {{<features>}}
 - When deploying an update, kotsadm operator removes old resources no longer exist in the new version
 - Added support for optional resources in KOTS applications. 
-This is specified using an annotation, documented https://kots.io/vendor/packaging/optional-resources/
+This is specified using an annotation, documented https://kots.io/vendor/packaging/include-resources/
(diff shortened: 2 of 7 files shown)


DanStough (approved):

DanStough:
Might benefit from @JustinN-Replicated reviewing from the product perspective.

emosbaugh (commented):
This doesnt feel like the right place to put this. Page title is `Optional Resources`.

tdensmore:
> This doesn't feel like the right place to put this. Page title is `Optional Resources`.

I can definitely be convinced. Do you think the page title is wrong, or should I put my addition of the `placeholder` annotation somewhere else? I only put it here because a Vendor said they thought it belonged here with the other annotations that drive conditional resources.

emosbaugh:
That and the intro paragraph detailing only when and exclude. I'll leave it to @JustinN-Replicated

JustinN-Replicated:
It seems like this article describes 'Kubernetes Resource Inclusivity' so maybe that is a more accurate title? Having said that, in the previous documentation it didn't look like we included an example of when a resource should be included either; could the same database example be used (IE if a customer did want to use embedded)?

tdensmore:
> It seems like this article describes 'Kubernetes Resource Inclusivity' so maybe that is a more accurate title? 

I just changed the filename to: `include-resources.md` and title to: "Including and Excluding Kubernetes Resources". does that work?

>Having said that, in the previous documentation it didn't look like we included an example of when a resource should be included either; could the same database example be used (IE if a customer did want to use embedded)?

I used the Ingress example because the `placeholder` annotation is currently mentioned [on that page](https://kots.io/vendor/packaging/ingress/). I wanted to keep everything consistent, but that might not be the best example to give? I'm open to changing it.

JustinN-Replicated:
@tdensmore  The new page title works for me - I think it is a bit more accurate and succinct enough.  

I am always +1 for consistency; unsure if there is an example for including we could add as well though I think that's not a deal breaker in my opinion. LGTM and let Wallaroo give us feedback?

tdensmore:
> @tdensmore The new page title works for me - I think it is a bit more accurate and succinct enough.
> 
> I am always +1 for consistency; unsure if there is an example for including we could add as well though I think that's not a deal breaker in my opinion. LGTM and let Wallaroo give us feedback?

Cool, thanks. I already ran this (on Slack) by Wallaroo and Mitch he approved and appreciated the improvements (will Slack you that thread) . If its okay - lets merge / release these changes and improve the example if needed?

JustinN-Replicated (approved):
LGTM

tdensmore:
Merging despite failure, since the failure happened because that new page doesn't exist yet.

PR #577, opened 2021-09-23
ignore-rollback flag for garbage-collect-images cli command

sgalsaleh:


DanStough (approved):
I feel like this flag needs some additional description of when it's OK to use, but I'm fine with the current changes.

divolgin (approved):

JustinN-Replicated:
We should mention the default (false); also may want to allude to impact (presumably that they cannot rollback or that it may impact the ability to rollback).

JustinN-Replicated:
Meant to take another look at this yesterday and incorporate feedback but you beat me to it @sgalsaleh - lgtm!

PR #584, opened 2021-10-06
docs: add info on beta helm chart features

morningvera:
[SH37064](https://app.shortcut.com/replicated/story/37064/provide-documentation-for-all-replicated-products-involved-with-the-usehelminstall-workflow)


JustinN-Replicated (approved):
Made some changes and in general looks good to me. Looping in @samantha-replicated and @jonquil2002

content/vendor/native-helm/using-native-helm-charts.md
```diff
@@ -0,0 +1,39 @@
...
+
+Native Helm hooks and weights enable more control over when resources are deployed. This is useful if you want to bundle actions as part of a release. For example, you can build in a database backup as part of the upgrade process while ensuring that the backup occurs prior to upgrading the rest of the resources. Helm weights provide even more control by governing the order of operations within each hook.
+
+The following hooks are currently supported:
+* pre-install - executes after resources are rendered but before any resources are installed.
+* post-install - executes after resources are installed.
+* pre-upgrade - executes after resources are rendered but before any resources are upgraded.
+* post-upgrade - executes after resources are upgraded.
+* pre-delete - executes before any resources are deleted.
+* post-delete - executes after resources are deleted.
```
JustinN-Replicated, line 37: We may need to be more explicit here as IIRC since we do not have the idea of a 'delete' flow, we will never execute these hooks. This is likely something we will need to address for GA I think, but right now I believe we take no action (I may be wrong).

morningvera: so is it fair to say that we don't actually support them?

JustinN-Replicated: It is a bit of a gray area, @jala-dx may be able to provide a bit more detail and context (I think since we can interoperate with them is why we deemed it 'supported' even though we do not take any action or anything with them).
I see this being problematic in GA though because if someone were to have these hooks, they would expect them to run (and I think they would, but never initiated by us - would always need to be initiated by a user who is running the helm delete command; where this dove tails with us is what we would need to change to recognize the delete - I'd imagine this is something like a chart being removed or modified or something? I am not 100%)

PR #603, opened 2021-11-04
Removed imaginary annotation called 'kots.io/include':  and replaced it with 'kots.io/when' 

tdensmore:
Removed imaginary annotation called 'kots.io/include':  and replaced it with 'kots.io/when'


samantha-replicated (approved):
LGTM!

PR #621, opened 2021-12-01
[1.58.0] release notes

sgalsaleh:


content/release-notes/1.58.0.md
```diff
@@ -0,0 +1,10 @@
+---
+date: 2021-12-01T08:00:00-05:00
+linktitle: "1.58.0"
+title: "1.58.0"
+kubernetes: "1.19, 1.20, and 1.21"
+---
+
+{{<features>}}
+ * Added support for [semantic version](https://semver.org/) ordering of releases when the version labels are valid [semantic versions](https://semver.org/). In order for this feature to be enabled, [semantic versioning must be enabled for the channel](/vendor/packaging/promoting-releases/#semantic-versioning) that the license is currently on.
```
ajp-io: It doesn't feel like we need to link this twice in the same sentence.
```suggestion
 * Added support for [semantic version](https://semver.org/) ordering of releases when the version labels are valid semantic versions. In order for this feature to be enabled, [semantic versioning must be enabled for the channel](/vendor/packaging/promoting-releases/#semantic-versioning) that the license is currently on.
```

PR #635, opened 2021-12-16
document the port number used for the admin-console command

BarryWilliams:
@replicatedhq/replicated-docs


tdensmore:
My only comment / caveat is that this port will be the default ONLY IF that port is free. If not, it will be a random number in the high range.

jonquil2002 (approved):
lgtm

samantha-replicated:
Found this hanging around, it was never merged. Merging now.

PR #642, opened 2022-01-07
Update helm-processing.md

JustinN-Replicated:
Re-organized to be more clear on which portions related to which helm install flow.


JustinN-Replicated:
@replicatedhq/replicated-docs (probably @jonquil2002 is closest to this) but based on [this comment](https://docs.google.com/document/d/1wGjoZ3AfKTtbIJexJwjbUk8JZgI9hHvM_oViiAG3ltk/edit?disco=AAAAT6zDPkA) I made this pr/changes so it was more clear. Wanted to start getting some eyes on it - will loop in @replicatedhq/day0  and @AmberAlston  as well

content/vendor/helm/helm-processing.md
```diff
@@ -113,4 +100,17 @@ Replicated leverages the Helm binary to install the fully-rendered chart resourc
...
+
+The Replicated Helm installation is compatible with both Helm v2 and Helm v3 as it does not rely on Tiller to deploy the chart into the cluster. Instead, KOTS treats a Helm Chart as the packaging spec rather than the deployment tool. The Replicated Helm installation creates deployable YAML by leveraging the same functionality that the `helm template` command uses, with some extended functionality for [specific Helm hooks](/vendor/packaging/cleaning-up-jobs/#helm-charts).
+
+When a [Helm chart based KOTS application is installed via airgap](/vendor/helm/helm-airgap-builder), the processing of the chart is managed in the end customer environment using either the KOTS CLI or the admin console. This means that the customer supplied values, license values, and existing values can be used to create the deployable manifests. 
+
+In either scenario, the resulting deployment is comprised of raw Kubernetes manifests. Therefore, cluster operators are always able to view the exact difference between what is currently deployed and what the update will deploy. This level of change management provides the necessary transparency to provide the full assurance that cluster operators require.
```
jonquil2002, line 111: "In either scenario" could be made clearer by specifying which scenarios. It's not really clear to me at least. Does this mean Helm v@ and v#? Something else, e.g. KOTS without embedded cluster and airgap?

content/vendor/helm/helm-processing.md
```diff
@@ -113,4 +100,17 @@ Replicated leverages the Helm binary to install the fully-rendered chart resourc
...
+
+### Replicated Helm Versioning Considerations
+To determine if Helm v3 is necessary, KOTS will check the apiVersion supplied in the `Chart.yaml` file of the Helm Chart. By default (if **Chart.yaml** is not supplied or apiVersion is not present), KOTS will use Helm V2 to process all Helm Charts to create deployable YAML. Optionally, an API version can be specified in the KOTS [`HelmChart`](https://kots.io/reference/v1beta1/helmchart/) resource. When the `helmVersion` property is set to `"v3"`, KOTS will use Helm v3 to process the Helm Chart.
+
+In the future KOTS may be updated to use Helm v3 by default.
```
jonquil2002: It's best not to talk about what might happen in the future. It doesn't really benefit the user bc it's speculation, and also can open legal ramifications.

jonquil2002 (dismissed):

jonquil2002:
@JustinN-Replicated I reviewed, made suggestions, and also made some specific commits. LGTM overall.

content/vendor/helm/helm-processing.md
JustinN-Replicated: I agree; I think the only reason this stayed in there prior to the last changes from @jeffreygolden was because it was existing. I, too, think it should be removed so will do that momentarily.

content/vendor/helm/helm-processing.md
JustinN-Replicated: I think it is 'in both airgap and non-airgap scenarios'  or 'in internet connected and airgap scenarios' 
I am unsure if there is a usual way we phrase this for 'connected' instances (IE those that are not airgap)

content/vendor/helm/helm-processing.md
JustinN-Replicated: from [here](/kotsadm/installing/online-install/) it looks like we refer to this as Online install so will adjust to reflect that here.

AmberAlston: I agree we should be removing any language we find in the docs where we are projecting into the future and there is no guarantee that change will actually happen.

morningvera (dismissed):

AmberAlston (approved):
@JustinN-Replicated With that latest commit, I'm good with this now. I'll approve.

PR #656, opened 2022-01-26
feat: add versioned kots install url

DanStough:


jonquil2002:
@DanStough  @ajp-io  Is this waiting on a Docs team review? Thanks.

ajp-io (approved):

ajp-io:
@jonquil2002 It is now! Just pushed a change for it.

jonquil2002 (approved):
lgtm

jonquil2002:
@ajp-io Review is done. You can go ahead and merge.

content/kots-cli/getting-started.md
```diff
@@ -21,11 +21,16 @@ If you're using an [embedded Kubernetes cluster](/kotsadm/installing/installing-
...
 
-To start, install KOTS using the Krew plugin manager:
+To start, install the latest version of KOTS using the Krew plugin manager:
 ```bash
 curl https://kots.io/install | bash
 ```
 
+You can also install a specific version of KOTS, like one that has been QA'ed by your team, using the versioned install URL:
+```bash
+curl https://kots.io/install/<version> | bash
```
markpundsack, line 31: Would be nice if there were an example here so you know the format of the version. e.g. with or without the `v` prefix.

ajp-io: It works either way, but I can do that.

ajp-io: I think originally it had a version in there, but I removed it since we'll never update the reference and people following the getting started guide shouldn't download an ancient version. But I'm sure there's a way I can word it to mitigate that.

PR #657, opened 2022-01-28
add k8s 1.23

stefanrepl:


hack/release-notes.md.tmpl
```diff
@@ -2,7 +2,7 @@
 date: __DATE_TIME__
 linktitle: "__VERSION__"
 title: "__VERSION__"
-kubernetes: "1.19, 1.20, and 1.21"
+kubernetes: "1.20, 1.21, 1.22, and 1.23"
```
sgalsaleh, line 5: we only list the latest 3 versions, so this should be
```suggestion
kubernetes: "1.21, 1.22, and 1.23"
```

stefanrepl: 1.20 is still seeing support as its EOL is in late feb.  This was discussed [here](https://replicated.slack.com/archives/CM8N4TWR0/p1643396382714039).  Can you review this thread, specifically what Mark and Alex said, and let me know if that changes your mind.

ajp-io: Yeah, we agreed to have all four versions in there since all four are still in support.

ajp-io (approved):
I'm approving, but we shouldn't merge this until KOTS 1.61 is released.

samantha-replicated (approved):
LGTM!

PR #661, opened 2022-02-02
builder is ONLY for determining images

laverya:


sgalsaleh (dismissed):

sgalsaleh (dismissed):

paigecalvert (approved):
Hi @laverya @sgalsaleh , just added a commit with some rewording--lmk if there's anything there that I need to change/has changed the technical meaning of your update.

PR #669, opened 2022-02-09
Add limitations for minKotsVersion

ajp-io:


content/reference/v1beta1/application.md
```diff
@@ -183,4 +183,6 @@ The minimum KOTS version that is required by the release.
...
+
+When promoting a new release that specifies a minimum KOTS version that is greater than what an end user currently has deployed, that release will not appear in the version history of the admin console after checking for updates. An error message will be displayed temporarily, but it will disappear after a few minutes. This error is also displayed when checking for updates using the [`kots upstream upgrade` command](/kots-cli/upstream/). End users must update their admin console to at least the minimum version in order to fetch the update without error. KOTS cannot update itself automatically, and there is no way to update KOTS from the admin console. When increasing the specified minimum KOTS version, vendors should tell their end users to download the appropriate version of the `kots` CLI and use the [`kots admin-console upgrade` command](/kots-cli/admin-console/upgrade/) to upgrade KOTS before retrieving the updated application version.
```
sgalsaleh, line 188: > When increasing the specified minimum KOTS version, vendors should tell their end users to download the appropriate version of the `kots` CLI and use the [`kots admin-console upgrade` command](/kots-cli/admin-console/upgrade/) to upgrade KOTS before retrieving the updated application version.

a couple of comments on this sentence:
* downloading the cli and running that command is only for existing cluster installs.
* should the vendor really tell their end users to upgrade kots? isn't that the purpose of the error message? end users should _refer_ to the vendor/vendor's docs for instructions on how to upgrade kots i feel.

ajp-io: 1. Good point, I'll make an adjustment to mention the existing and embedded cluster options.
2. Yes, that's the purpose of the error message, but the error message is fleeting. If someone has auto check for updates enabled, they might never see that message. Seems like the vendor should make sure their end users know about the need to update until we can have a clearer way of communicating the need to upgrade in the admin console (or updating KOTS automatically).

sgalsaleh (dismissed):

ajp-io:
@replicatedhq/replicated-docs Added some more words on the limitations of this recent beta feature. Ready for review.

paigecalvert:
Thanks @ajp-io --just added a commit with some edits and formatting updates. Was hoping to break up the paragraphs a bit so that it was easier to scan for the important info. LMK if this look alright to you.

Not strictly part of this PR, but I also noticed what appears to be a missing word in the paragraph above:
>...for both new installations and application updates, **and will block if** the currently deployed KOTS version...

Should this say: "and will block installations"? or maybe "and will block installations and updates?"

ajp-io:
@paigecalvert I made a change to address your comment. I also changed wording a bit. Can you let me know if it looks okay?

paigecalvert (approved):
Yep LGTM @ajp-io !

PR #671, opened 2022-02-11
[v1.63.0] Adds release notes for kots

trelore:


ajp-io (approved):

content/release-notes/1.63.0.md
```diff
@@ -0,0 +1,14 @@
...
+title: "1.63.0"
+kubernetes: "1.20, 1.21, 1.22, and 1.23"
+---
+
+{{<features>}}
+* Changes the [`kots upstream upgrade`](/kots-cli/upstream/) command to be synchronous by default and exposes error messages for it.
+{{</features>}}
+
+{{<changes>}}
+* Sets the Native Helm timeout to 60 minutes instead of the default 5 minutes.
```
paigecalvert, line 13: Question: does this mean that 60 minutes is the new default? 

If so, recommend adding the word default to the first part to clarify: "Sets the native Helm default timeout to 60 minutes instead of 5 minutes"

ajp-io: Jeff would know best, but I read this as, Helm defaults to five minutes, so that's what our Native Helm support was using. We're now overriding the default to use 60 minutes instead. So the original wording makes sense to me, but if Paige didn't follow, then maybe it's not quite clear! I agree with Alexander that the word default could be dropped and it wouldn't make a difference, so if that makes it clearer, let's do that.

paigecalvert: Yep totally fair--up to you all! I think it was just that "instead of the default 5 minutes" that threw me off.

I wondered: "what is the significance of using the word default when referring to the 5 min, and not when referring to the 60 minutes?"

ajp-io (approved):

PR #676, opened 2022-02-24
Say which KOTS version introed min and target KOTS version

ajp-io:


ajp-io:
@replicatedhq/replicated-docs Ready for your reivew.

paigecalvert (approved):

PR #677, opened 2022-02-25
[1.65.0] Kots upstream download CLI command docs

sgalsaleh:
## Screenshots:

Image 2022-02-24 at 5 03 50 PM

Image 2022-02-24 at 5 03 56 PM


content/kots-cli/upstream/download.md
```diff
@@ -0,0 +1,33 @@
...
+
+| Flag                              | Type   | Description                                                                                      |
+|:----------------------------------|--------|--------------------------------------------------------------------------------------------------|
+| `-h, --help`                      |        | help for upstream download                                                                       |
+| `--kubeconfig`                    | string | the kubeconfig to use _(default is `$KUBECONFIG`. If unset, then `$HOME/.kube/config`)_          |
+| `-n, --namespace`                 | string | the namespace where the admin console is running _(required)_                                    |
+| `--sequence`                      | int    | The local app sequence for the version to retry downloading _(required)_                         |
+| `--skip-preflights`               | bool   | set to true to skip preflight checks                                                             |
+| `--skip-compatibility-check`      | bool   | set to true to skip compatibility checks between the current kots version and the update         |
```
ajp-io, line 26: Can you explain what this one does? Does this somehow ignore the `minKotsVersion` and download it anyway?

content/kots-cli/upstream/download.md
```diff
@@ -0,0 +1,33 @@
...
+```
+* _Replace `[app-slug]` with the app slug for your KOTS application (required)._
+* _Provide `[flags]` according to the table below._
+
+| Flag                              | Type   | Description                                                                                      |
+|:----------------------------------|--------|--------------------------------------------------------------------------------------------------|
+| `-h, --help`                      |        | help for upstream download                                                                       |
+| `--kubeconfig`                    | string | the kubeconfig to use _(default is `$KUBECONFIG`. If unset, then `$HOME/.kube/config`)_          |
+| `-n, --namespace`                 | string | the namespace where the admin console is running _(required)_                                    |
+| `--sequence`                      | int    | The local app sequence for the version to retry downloading _(required)_                         |
```
ajp-io, line 24: Will `upstream upgrade` print out the sequence? So that if it errors, someone can get the sequence and bring it right over to this `download` command.

ajp-io (commented):
Made a few tiny changes and left two comments. Thanks for updating all those links to be relative instead of absolute too. That's a good change.

content/kots-cli/upstream/download.md
sgalsaleh: Yes, the same flag exists for the "upstream upgrade" command.

sgalsaleh: We wanted an escape hatch, but we didn't want to make it easy, so we only added this flag to the CLI, and there's no way to do it from the UI.

content/kots-cli/upstream/download.md
sgalsaleh: Good question, it will not, but we can make it do that.

ajp-io: I don't want to bloat the feature, but I'm concerned `download` is useless if the sequence isn't available. Something to think about. Maybe there's another way it's available, but in the `upstream upgrade` output would be the best UX since they probably go to `download` after running `upgrade`.

ajp-io:
@replicatedhq/replicated-docs This is ready for review. I didn't touch the CLI parameters stuff because I had a feeling you would have an opinion on how that should be formatted (capitalization, periods, etc.). So feel free to make that how you think it should be.

content/kots-cli/upstream/download.md
ajp-io: Decided in standup to leave this is as since other commands require the sequence and provide no way to know what it is either. It's a bad experience, but it's no different than other commands at this point.

paigecalvert (approved):
LGTM! Added a few commit with edits.

I did remove Mikael's handle--best to keep any personal info out of the docs, even in the form of a shoutout.

PR #706, opened 2022-05-06
45222 removes old content from this repo

paigecalvert:
This PR is the last step in the migration of content from kots.io to replicated-docs. 

To close out the migration, we want to remove the `/content/` and `/static/` folders from kots.io, which contain now out-dated content for Replicated. The goal of removing this content is to help avoid duplication and any possible confusion.

We originally thought about doing this a couple months ago after we completed the migration, but decided it would make sense to keep the content around for awhile to make sure it wasn't needed for anything. We haven't missed it, so away it goes.

https://app.shortcut.com/replicated/story/45222/first-week-of-may-delete-the-content-from-kots-io-repo


GraysonNull (dismissed):
I think this all looks fine. I ran it locally as well and it seemed ok. Assuming we have all the appropriate redirects in place I think this is good to go

GraysonNull (approved):

